PR #1, opened 2017-05-04
Add support for pausing jobs in the Executor

haveo:


core/src/main/scala/com/criteo/cuttle/Execution.scala
```diff
@@ -49,35 +51,85 @@ object ExecutionPlatform {
...
-              nextExecutionId,
-              job.id,
-              startTime,
-              LocalDateTime.now(),
-              context.toJson,
-              result.isSuccess
-            ))
-          .transact(xa)
-          .unsafePerformIO
+    val maybeFuture: Option[Future[Unit]] = atomic { implicit txn =>
```
guillaumebort: maybePausedExecution

PR #2, opened 2017-05-04
TimeSeries: refactor support for backfills

haveo:


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeseriesScheduler.scala
```diff
@@ -28,7 +32,7 @@ case class Daily(tz: ZoneId) extends TimeSeriesGrid
 
 private case object Continuous extends TimeSeriesGrid
 
-case class Backfill(id: Int,
+case class Backfill( //id: Int,
```
guillaumebort: ?

PR #10, opened 2017-05-15
Throttle execution retries.

guillaumebort:
We define a retry as an execution (job, context) that fail and is followed
by a new attempt to execute it. When it happens, instead of just executing
again right away, we throttle it based on a RetryStrategy function. The default
strategy implements an exponential backoff retry delay.

It also allow us to define a list of failing executions that will appear in the
UI, and we define a new API endpoint for it.

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Execution.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Execution.scala
@@ -1,6 +1,7 @@
 package com.criteo.cuttle
 
-import java.time.{LocalDateTime, ZonedDateTime}
+import java.util.{Timer, TimerTask}
+import java.time.{Duration, LocalDateTime, ZoneId, ZonedDateTime}
 import java.time.format.DateTimeFormatter.{ISO_INSTANT}
 
 import scala.util.{Failure, Success}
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/criteo/cuttle/Execution.scala
```diff
@@ -174,8 +203,26 @@ case class Executor[S <: Scheduling](platforms: Seq[ExecutionPlatform[S]], queri
...
           case Success(()) =>
             execution.streams.debug(s"Execution successful.")
+            atomic { implicit tx =>
+              recentFailures -= (execution.job -> execution.context)
+            }
           case Failure(e) =>
             execution.streams.debug(s"Execution failed with message: ${e.getMessage}.")
+            atomic {
+              implicit tx =>
+                recentFailures --= recentFailures
```
haveo: `recentFailures.retain`

core/src/main/scala/com/criteo/cuttle/Execution.scala
```diff
@@ -174,8 +203,26 @@ case class Executor[S <: Scheduling](platforms: Seq[ExecutionPlatform[S]], queri
         .andThen {
           case Success(()) =>
             execution.streams.debug(s"Execution successful.")
+            atomic { implicit tx =>
+              recentFailures -= (execution.job -> execution.context)
```
haveo: `recentFailures.single -= (execution.job -> execution.context)`

PR #18, opened 2017-05-22
Move MySQL storage and querying logic for contexts to the scheduler

haveo:
The `executions` table now contains a `context_id` column instead of a
JSON representation of contexts. The scheduler provides an SQL query to
translate these ids into JSON. It is also possible to provide a custom
SQL query to get execution logs for only a subset of all contexts.

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Database.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Database.scala
@@ -43,11 +43,12 @@ object Database {
         job         VARCHAR(1000) NOT NULL,
         start_time  DATETIME NOT NULL,
         end_time    DATETIME NOT NULL,
-        context     JSON NOT NULL,
+        context_id  VARCHAR(1000) NOT NULL,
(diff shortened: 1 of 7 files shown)


core/src/main/scala/com/criteo/cuttle/Database.scala
```diff
@@ -43,11 +43,12 @@ object Database {
         job         VARCHAR(1000) NOT NULL,
         start_time  DATETIME NOT NULL,
         end_time    DATETIME NOT NULL,
-        context     JSON NOT NULL,
+        context_id  VARCHAR(1000) NOT NULL,
```
vguerci: `VARCHAR(1000)`? Why not an uuid `CHAR(36)` like `executions.id`?

haveo: It's not really an opaque ID like a GUUID it's also meant to encode part of the context so the executor can use it as a sorting key when listing executions.

PR #20, opened 2017-05-23
Import scala.language.postfixOps

jbkt:
Fix compilation warning.


haveo:
I'd rather not use postfix ops. I fixed the warning in one of my PRs by changing `1 second` to `1.second`.

PR #21, opened 2017-05-23
Fix date computation

jbkt:


guillaumebort:
Ok let's use an `Instant` for the start parameter then. So you can use an `Instant` in the example.

jbkt:
Done

timeseries/src/main/scala/com/criteo/cuttle/timeseries/package.scala
```diff
@@ -1,26 +1,26 @@
...
-    def date(args: Any*): LocalDateTime = macro safeLiteralDate
+    def date(args: Any*): Instant = macro safeLiteralDate
   }
 
-  def safeLiteralDate(c: Context)(args: c.Expr[Any]*): c.Expr[LocalDateTime] = {
+  def safeLiteralDate(c: blackbox.Context)(args: c.Expr[Any]*): c.Expr[Instant] = {
     import c.universe._
     c.prefix.tree match {
       case Apply(_, List(Apply(_, Literal(Constant(dateString: String)) :: Nil))) =>
         scala.util.Try(ZonedDateTime.parse(dateString)) match {
```
guillaumebort: This has to be changed accordingly.

jbkt: Done

PR #48, opened 2017-06-13
List executions for a given period.

guillaumebort:


guillaumebort:
Yes I used it as a shortcut during the summit :)

PR #49, opened 2017-06-13
eslint tweaks

vguerci:
Note that we have ~180 errors/warns we should care about one day :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,9 @@
     "whatwg-fetch": "^2.0.3"
   },
   "scripts": {
-    "start": "./node_modules/devloop/bin/main.js",
-    "format": "./node_modules/prettier-eslint-cli/dist/index.js \"**/src/**/*.js\" -l info --write --bracket-spacing true"
+    "start": "loop",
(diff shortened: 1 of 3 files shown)


guillaumebort:
I thought that prettier-eslint should take care of fixing it automatically?

vguerci:
`prettier-eslint` indeed calls `eslint --fix` but [not all rules have automatic fix](http://eslint.org/docs/rules/)

side note, many easy manual fix, like (~100) errors coming from lodash rules like:
```warning  Prefer '_.map' over the native function                lodash/prefer-lodash-method```
not sure what are the best practices with these, we could also set off these.

Also, IntelliJ shows eslint messages in editor, vim and atom can as well I think.

File: package.json [hunk 1]
```diff
@@ -64,7 +64,9 @@
     "whatwg-fetch": "^2.0.3"
   },
   "scripts": {
-    "start": "./node_modules/devloop/bin/main.js",
```
guillaumebort: Sorry I missed that during the review but it is better to use the local version of devloop. Why did you change it? I think it should be reverted.

PR #54, opened 2017-06-14
[backfills] wip

vguerci:
This is an ugly piece of code that kinda works to create backfills, not intended for merge, but to review & discuss.

### DIFF
--- a/core/src/main/javascript/app/pages/BackfillCreate.js
+++ b/core/src/main/javascript/app/pages/BackfillCreate.js
@@ -3,7 +3,7 @@
 import React from "react";
 import { connect } from "react-redux";
 import injectSheet from "react-jss";
-import goBack from "redux-url";
+import { goBack } from "redux-url";
 import CloseIcon from "react-icons/lib/md/close";
 import moment from "moment";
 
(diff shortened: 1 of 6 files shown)


core/src/main/javascript/app/pages/BackfillCreate.js
```diff
@@ -0,0 +1,223 @@
...
+    let { classes, workflow, back } = this.props;
+    let { selectedJobs } = this.state;
+
+    return (
+      <Window title="Create Backfill">
+        <CloseIcon className={classes.close} onClick={back} />
+        <form onSubmit={this.createBackfill}>
+          <FancyTable key="properties">
+            <dt key="name_">Name:</dt>
+            <dd key="name"><input ref="_name" type="text" required /></dd>
```
guillaumebort: I don't think you need to keep a reference to these form elements. The state of the main component should contain their data, and you can you onChange event handlers to handle state changes.

core/src/main/javascript/app/pages/BackfillCreate.js
```diff
@@ -0,0 +1,223 @@
...
+            }
+            onChange={this.selectJobs}
+          />
+        </div>
+      </Window>
+    );
+  }
+}
+
+// TODO duplicated styles
```
guillaumebort: Yes I'm not sure what is this style for :)

core/src/main/javascript/app/pages/BackfillCreate.js
vguerci: done, also improved a bit using HTML5 datetime inputs, will change this to a text input

core/src/main/javascript/app/pages/BackfillCreate.js
vguerci: Cleaned up styles, but still some duplication / hacks

vguerci:
Abandoned, see #65 instead.

PR #57, opened 2017-06-15
Eslint: fix and adjust rules

vguerci:
No more errors, could be useful to make eslint break builds, not sure of what is a good way to achieve this, through sbt call `yarn eslint` or maybe invoking from webpack

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,7 +24,8 @@
   "parser": "babel-eslint",
   "env": {
     "es6": true,
-    "node": true
+    "node": true,
+    "browser": true
   },
   "rules": {
     "prettier/prettier": "warn",
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,6 @@
     "eslint:recommended",
     "plugin:react/recommended",
     "plugin:flowtype/recommended",
-    //"plugin:lodash/recommended",
     "prettier",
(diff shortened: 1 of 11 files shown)


core/src/main/javascript/app/components/JobSelector.js
```diff
@@ -292,7 +292,6 @@ const styles = {
       }
     },
     "& .Select-value": {
-      lineHeight: "1em",
```
guillaumebort: Are you sure of that? Was it a warning from eslint? Because it could override another rule.

.eslintrc
```diff
@@ -9,7 +9,7 @@
     "eslint:recommended",
     "plugin:react/recommended",
     "plugin:flowtype/recommended",
-    "plugin:lodash/recommended",
+    //"plugin:lodash/recommended",
```
guillaumebort: Can we just remove it?

core/src/main/javascript/app/components/JobSelector.js
vguerci: I am, the error was duplicate, see 5 lines below `lineHeight: "18px",`

.eslintrc
vguerci: Sure

PR #72, opened 2017-06-19
Add details for jobs in the workflow screen

hhalex:


core/src/main/javascript/datamodel.js
```diff
@@ -43,10 +43,28 @@ export type Tag = { name: string, description: string };
...
+      grid: {
+        period: "daily",
+        zoneId: string
+      },
+      start: string,
+      maxPeriods: number
+    }
+  | {
+      grid: {
+        period: "hourly" | "continuuous"
```
vguerci: typoo ;p

PR #79, opened 2017-06-21
Creates a basic authentication mechanism

dufrannea:
Authenticator trait and two implementations have been introduced :
- NoAuth : logs everybody as guests
- HttpBasicAuth : logs following http basic auth mechanism


core/src/main/scala/com/criteo/cuttle/Authentication.scala
```diff
@@ -0,0 +1,89 @@
...
+    authenticate(request)
+      .fold(user => wrappedAuthedService(user)(request), identity)
+  }
+
+  /**
+    * Performs authentication on request.
+    * @param r the request to be authenticated
+    * @return either an authenticated user or a response.
+    */
+  def authenticate(r: Request): Either[User, Response]
```
guillaumebort: Just a matter of style: we usually use Right for the successful value and in this case it is most likely the User. So I would inverse both types.

core/src/main/scala/com/criteo/cuttle/Authentication.scala
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+  * Implementation of the HTTP Basic auth.
+  * @param credentialsValidator method to validate credentials.
+  */
+case class BasicAuth(credentialsValidator : ((String,String)) => Boolean) extends Authenticator {
+
+  val scheme = "Basic"
+  val unauthorizedResponse = Response(401).addHeaders(h"WWW-Authenticate" -> h"""Basic realm="User Visible Realm"""")
```
guillaumebort: Please change the realm for something more relevant (or make it configurable with a default value in the BasicAuth constructor).

dufrannea:
@guillaumebort Could you validate the new design ?

PR #82, opened 2017-06-22
Bootstrap Monotonic scheduler

sdjamaa:
Add first steps of the Monotonic scheduler.
It's currently running as it is but there's no state management yet.

### DIFF
--- a/monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
+++ b/monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
@@ -1,0 +1,164 @@
+package com.criteo.cuttle.monotonic
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+import com.criteo.cuttle.{Executor, Job, Scheduler, Scheduling, SchedulingContext, Workflow, XA, utils}
+import doobie.imports._
+import io.circe.{Encoder, Json}
+
+import scala.concurrent.Future
+import scala.concurrent.duration.{Duration => ScalaDuration}
+import scala.concurrent.stm._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.reflect.ClassTag
+
+object MonotonicContext {
+  def apply[T: Encoder: ClassTag](execParams: T, execTime: LocalDateTime) = new MonotonicContext[T] {
+    def executionParams = execParams
+
+    def executionTime = execTime
+
+    val enc = implicitly[Encoder[T]]
+
+    override def toJson: Json = Json.obj(
+      "exec_params" -> enc(execParams),
+      "exec_time" -> Json.fromString(DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(executionTime))
+    )
+
+    override def log: ConnectionIO[String] = Database.serializeContext(this)
+
+    override def compareTo(other: SchedulingContext): Int = other match {
+      case o: MonotonicContext[T] => ordering.compare(this, o)
+    }
+  }
+
+  implicit def ordering[T] = Ordering.fromLessThan((t1: MonotonicContext[T], t2: MonotonicContext[T]) => t1.executionTime.isBefore(t2.executionTime))
+}
+
+trait MonotonicContext[T] extends SchedulingContext {
+  /**
+    * Parameters of jobs.
+    * Each graph execution will provide different values.
+    *
+    * @return
+    */
+  def executionParams: T
+
+  /**
+    * Workflow execution time, used to order executions.
+    * This is the time of the completion of the trigger.
+    *
+    * @return graph execution time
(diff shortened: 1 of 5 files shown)


monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  def enc: Encoder[T]
+}
+
+class TriggerJob[T](triggerName: String, trigger: () => Future[T], mon: MonotonicScheduling[T])
+  extends Job[MonotonicScheduling[T]](id = triggerName, name = "trigger_job", description = "Triggers the execution of a workflow", tags = Set.empty, scheduling = mon)(_ => Future.successful(())) {
+  def apply() = trigger()
+}
+
+case class MonotonicScheduling[T: ClassTag: Encoder](trigger: MonotonicScheduling.Aux[T]) extends Scheduling {
```
guillaumebort: I think there is a problem in the design here. If I understand well the trigger function is the one that will trigger the whole workflow execution. Because of that it should be a property of the scheduler and not of the scheduling. The scheduling only allow to configure job related properties.

I propose this change:
- `MonotonicScheduling` is an object instead (as for now there is no job dependant configuration).
- The trigger is defined as an implicit value in the user code, and is used to automatically provide an implicit scheduler value for Monotonic.

Also I understand how the trigger function will run a workflow but perhaps it would be better with a more imperative API where user code explicitly trigger a run. For example, assuming there is no trigger function anymore:

```
val world = Job("world", Monotonic) { implicit exec =>
  ???
}

val hello = Job("hello", scheduling = monotonicScheduling) { implicit exec =>
  ???
}

val project = CuttleProject("Hello World") {
  world dependsOn hello
}

project.start()

(1 to 10).map(i => project.scheduler.trigger(i))
```

I say that because it is usually easier for user code to write this kind of imperative code than a trigger function that will loop. Also it will be easier to map to a model where you can trigger a run via the HTTP API. And triggering based on an event queue like kafka is as easy.

monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
```diff
@@ -0,0 +1,165 @@
...
+        state.get(context).getOrElse(Set.empty) + job)
+    }
+
+
+
+  def start(graph: Workflow[MonotonicScheduling[T]], executor: Executor[MonotonicScheduling[T]], xa: XA) = {
+    Database.doSchemaUpdates.transact(xa).unsafePerformIO
+
+    def runTrigger(): Unit =
+      trigger() onSuccess { case res =>
```
guillaumebort: In the current implementation, a failure of the trigger function basically kill the scheduler forever as the failure is not handled.

monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
```diff
@@ -0,0 +1,165 @@
...
+    */
+  def executionParams: T
+
+  /**
+    * Workflow execution time, used to order executions.
+    * This is the time of the completion of the trigger.
+    *
+    * @return graph execution time
+    */
+  def executionTime: LocalDateTime
```
guillaumebort: Can't we use a UUID instead? It's unlikely that 2 workflow start at the same date but still it is kind of strange to use the start time as identifier...

sdjamaa: I'm using this to order executions and recreate a workflow from that execution time (to browse history of runs). This is the execution time of a workflow, not a job (I'll rename the func). I can add a UUID to act as a primary key but it may be useless.

monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
sdjamaa: It was a property of the scheduler but I had issues with typing and generics... That's why I moved everything to Scheduling. I'll try to solve the issues with typing and get back to you if I get stuck.

Users don't handle any loop, it's handled by the code. But yes they still need to provide a function that need to handle failures and retries (for example: I trigger the execution of a workflow when data is retrieved from an API, if there's no data I continue to query the API).

I'll see how to make it "mandatory" and maybe provide a default implementation (a timer).

monotonic/src/main/scala/com/criteo/cuttle/monotonic/MonotonicScheduler.scala
sdjamaa: Yep, I didn't handle error cases now and I also need to add unit tests.

haveo:
Closed as inactive.

PR #86, opened 2017-06-26
[backfills] Add Table and create/update backfills persistence

vguerci:
Basic queryBackfills to read table will be reworked when working on UI with paging

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
@@ -34,10 +34,11 @@ object TimeSeriesGrid {
   implicit val gridEncoder = new Encoder[TimeSeriesGrid] {
     override def apply(grid: TimeSeriesGrid) = grid match {
       case Hourly => Json.obj("period" -> "hourly".asJson)
-      case Daily(tz: ZoneId) => Json.obj(
-        "period" -> "daily".asJson,
-        "zoneId" -> tz.getId().asJson
-      )
+      case Daily(tz: ZoneId) =>
+        Json.obj(
+          "period" -> "daily".asJson,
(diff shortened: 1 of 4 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -140,7 +145,11 @@ case class TimeSeriesScheduler() extends Scheduler[TimeSeries] with TimeSeriesAp
           job -> (_state().apply(job) - Interval.closedOpen(start, end)))
         Right(id)
       }
+      (result, newBackfill)
     }
+    Database.createBackfill(newBackfill).transact(xa).unsafePerformIO
```
haveo: You shouldn't log the backfill into the DB if it's invalid (partial periods, etc.)

vguerci: Done, my bad, rebased naively w/o reading changes ;p

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -150,11 +150,11 @@ case class TimeSeriesScheduler() extends Scheduler[TimeSeries] with TimeSeriesAp
         _backfills += newBackfill
         _state() = _state() ++ jobs.map((job: TimeSeriesJob) =>
           job -> (_state().apply(job) - Interval.closedOpen(start, end)))
+        Database.createBackfill(newBackfill).transact(xa).unsafePerformIO
```
haveo: You can't do IO inside an STM transaction. Just use `result` like
`result.right.foreach (_ => Database.createBackfill(newBackfill).transact(xa).unsafePerformIO)`

PR #88, opened 2017-06-27
Better navigation interface for focused calendar

hhalex:


hhalex:
Agreed, added window resizer

PR #95, opened 2017-07-03
TimeSeries: deprecate continuum (replaced by IntervalMap)

haveo:
Instead of using the IntervalSet of continuum we now use our own
IntervalMap which allows us to tag intervals with different values.


guillaumebort:
Please put everything related to the IntervalMap `private[timeseries]`. You can generate the documentation with `unidoc` in sbt to check that nothing useless surface into the user API.

haveo:
`scaladoc` should be good now.

PR #118, opened 2017-07-10
Adds link highlighting in job logs

dufrannea:
Uses https://github.com/gregjacobs/Autolinker.js/

### DIFF
--- a/core/src/main/javascript/app/utils/URLHighlighter.js
+++ b/core/src/main/javascript/app/utils/URLHighlighter.js
@@ -1,0 +1,51 @@
+// @flow
+
+import React from "react";
+import AutoLinker from "autolinker";
+
+import classNames from "classnames";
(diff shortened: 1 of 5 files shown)


core/src/main/javascript/app/utils/URLHighlighter.js
```diff
@@ -0,0 +1,51 @@
...
+
+type Match = {
+    getType(): string,
+    getMatchedText(): string,
+    getOffset(): number,
+    getUrl(): string
+};
+
+const styles = {
+    color: 'lightcyan'
```
guillaumebort: Can you use color: 'inherit' instead?

guillaumebort: Can you use color: 'inherit' instead?

PR #128, opened 2017-07-25
Add user info to logs & backfills

dufrannea:


timeseries/src/main/scala/com/criteo/cuttle/timeseries/Database.scala
```diff
@@ -54,6 +54,11 @@ private[timeseries] object Database {
 
       CREATE INDEX timeseries_backfills_by_date ON timeseries_backfills (created_at);
       CREATE INDEX timeseries_backfills_by_status ON timeseries_backfills (status);
+    """.update,
+
+    sql"""
```
guillaumebort: Please just change the initial CREATE TABLE statement for now.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/Database.scala
```diff
@@ -48,17 +48,13 @@ private[timeseries] object Database {
         start       DATETIME NOT NULL,
         end         DATETIME NOT NULL,
         created_at  DATETIME NOT NULL,
+        created_by VARCHAR(100) NOT NULL,
```
vguerci: omg, missing space there, balance is broken ;p

PR #132, opened 2017-07-26
Monitors waiting time for executions

dufrannea:
- Monitors via executor polling
- Adds waitingSeconds info to execution model
- Adds a progressbar component to display waiting / running times


core/src/main/scala/com/criteo/cuttle/Cuttle.scala
```diff
@@ -20,6 +20,9 @@ class CuttleProject[S <: Scheduling] private[cuttle] (
   ) = {
     val xa = Database.connect(databaseConfig)
     val executor = new Executor[S](platforms, xa)(retryStrategy)
+
+    WaitingTimesMonitor.start(executor, xa)
```
guillaumebort: Its should be automatically started on Executor create. The whole WaitingTimes monitoring stuff should not leak outside of Executor.

core/src/main/scala/com/criteo/cuttle/WaitingTimesMonitor.scala
```diff
@@ -0,0 +1,18 @@
+package com.criteo.cuttle
+
+import scala.concurrent.duration._
+
+object WaitingTimesMonitor extends Queries {
```
guillaumebort: No need to extends Queries

guillaumebort: Also this code should be embeded into Executor. It's an implementation detail of the Executor.

PR #137, opened 2017-08-01
Marks retrying jobs as started

dufrannea:


core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -82,6 +81,8 @@ case class Execution[S <: Scheduling](
   def isCancelled = cancelSignal.isCompleted
   val cancelled = cancelSignal.future
   var waitingSeconds = 0
```
guillaumebort: Mutable state here should be private.

PR #140, opened 2017-08-03
Fix sort order in 'Finished'

Lordshinjo:


dufrannea:
ASC/DESC applies to the preceding column, and ASC is the default sort order. Repeating ASC is not necessary, we could actually remove them for consistence.

PR #150, opened 2017-08-21
Adds db connectivity check to status

dufrannea:


core/src/main/scala/com/criteo/cuttle/App.scala
```diff
@@ -142,12 +142,16 @@ private[cuttle] case class App[S <: Scheduling](project: CuttleProject[S], execu
...
     case GET at url"/api/status" =>
-      Ok(
-        Json.obj(
-          "project" -> project.name.asJson,
-          "version" -> Option(project.version).filterNot(_.isEmpty).asJson,
-          "status" -> "ok".asJson
-        ))
+      executor.healthCheck
+        .fold(
+          t => InternalServerError(t.getMessage()),
```
guillaumebort: why not returning the same JSON structure in case of error?

PR #153, opened 2017-08-25
Minor refactorings

dufrannea:
- Moves backfills to a Ref[Set] rather than TSet
- Extracts pure state transitions functions
- Fixes a few warnings

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
@@ -295,7 +295,7 @@ case class TimeSeriesScheduler() extends Scheduler[TimeSeries] with TimeSeriesAp
       val now = Instant.now
       val (stateSnapshot, completedBackfills, toRun) = atomic { implicit txn =>
         val (stateSnapshot, newBackfills, completedBackfills) =
-          computeNextState(workflow, now, _state(), _backfills(), completed)
+          collectCompletedJobs(workflow, now, _state(), _backfills(), completed)
 
         val toRun = jobsToRun(workflow, stateSnapshot, now)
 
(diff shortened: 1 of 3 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -287,66 +287,91 @@ case class TimeSeriesScheduler() extends Scheduler[TimeSeries] with TimeSeriesAp
...
       val now = Instant.now
       val (stateSnapshot, completedBackfills, toRun) = atomic { implicit txn =>
-        completed.foreach {
-          case (job, context, future) =>
-            val jobState = if (future.value.get.isSuccess) Done else Todo(context.backfill)
-            _state() = _state() + (job ->
-              (_state().apply(job).update(context.toInterval, jobState)))
-        }
+        val (stateSnapshot, newBackfills, completedBackfills) =
+          computeNextState(workflow, now, _state(), _backfills(), completed)
```
haveo: I think that name might be confusing because the state is still being updated using `jobsToRun`. Maybe something like `collectCompletedJobs` would be better?

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -287,66 +287,91 @@ case class TimeSeriesScheduler() extends Scheduler[TimeSeries] with TimeSeriesAp
...
 
       val newExecutions = executor.runAll(toRun)
 
       atomic { implicit txn =>
-        _state() = newExecutions.foldLeft(_state()) { (st, x) =>
-          val (execution, _) = x
-          st + (execution.job ->
-            st(execution.job).update(execution.context.toInterval, Running(execution.id)))
+        _state() = newExecutions.foldLeft(_state()) {
+          case (st, (execution, _)) =>
```
haveo: Apparently this works.
 http://scala-lang.org/files/archive/spec/2.11/08-pattern-matching.html#pattern-matching-anonymous-functions
I did not know that.

PR #163, opened 2017-09-08
Add execution.synchronize helper

guillaumebort:


examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
```diff
@@ -23,7 +23,7 @@ object HelloWorld {
       }
 
     val hello2 =
-      Job("hel.lo2", hourly(start), "Hello 2") { implicit e =>
+      Job("hello2", hourly(start), "Hello 2") { implicit e =>
```
dufrannea: This has been introduced intentionally as part of https://github.com/criteo/cuttle/pull/148 in order to reproduce a bug with ids containing dots.

guillaumebort: I know but I don't think the examples should be used for that. Better to introduce test projects in `src/test/` for that.

dufrannea: Agreed

PR #170, opened 2017-09-20
Better cancel listener design

guillaumebort:
To avoid memory leak for long execution registering several cancel listeners,
we now allow to unsubscribe when the code is not anymore interested by the cancel
signal.

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Executor.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Executor.scala
@@ -71,7 +71,11 @@ private[cuttle] class ExecutionStat(
    val status : ExecutionStatus
 )
 
-private object ExecutionCancelledException extends RuntimeException("Execution cancelled")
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -81,23 +85,49 @@ case class Execution[S <: Scheduling](
...
+        val listeners = cancelListeners.snapshot
+        cancelListeners.clear()
+        (true, listeners)
+      }
+    }
+    if(hasBeenCancelled) {
       streams.debug(s"Execution has been cancelled by user ${user.userId}.")
-      true
-    } else false
+      listeners.foreach(_.thunk())
```
dufrannea: Just curious if a thunk fails should the other be triggered ?

guillaumebort: Fixed

PR #175, opened 2017-09-27
Add the endpoint /api/metrics for Prometheus.

eryshev:
Add the endpoint /api/metrics for Prometheus.
Delete a logging package.
Small refactoring in Executor.scala.

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Executor.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Executor.scala
@@ -13,14 +13,15 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.stm._
 import scala.concurrent.stm.Txn.ExternalDecider
 import scala.concurrent.duration._
-import scala.reflect.{classTag, ClassTag}
+import scala.reflect.{ClassTag, classTag}
 import lol.http.PartialService
 import doobie.imports._
 import cats.implicits._
 import io.circe._
 import io.circe.syntax._
-
 import authentication._
+
+import Metrics._
 
 trait RetryStrategy {
   def apply[S <: Scheduling](job: Job[S], context: S#Context, previouslyFailing: List[String]): Duration
@@ -390,7 +391,7 @@ class Executor[S <: Scheduling] private[cuttle] (
     ).asJson
   }
 
-  override private[cuttle] def getMetrics(jobs: Set[String]): Seq[Metric] = {
+  override def getMetrics(jobs: Set[String]): Seq[Metric] = {
     val (running, waiting) = runningExecutionsSizes(jobs)
 
     Seq(
--- a/core/src/main/scala/com/criteo/cuttle/Executor.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Executor.scala
@@ -377,10 +377,18 @@ class Executor[S <: Scheduling] private[cuttle] (
                                          limit: Int): Seq[ExecutionLog] =
     queries.getExecutionLog(queryContexts, jobs, sort, asc, offset, limit).transact(xa).unsafePerformIO
 
+  /**
+    * Atomically get executor stats.
+    * @param jobs the list of jobs ids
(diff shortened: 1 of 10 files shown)


core/src/main/scala/com/criteo/cuttle/Metrics.scala
```diff
@@ -0,0 +1,22 @@
+package com.criteo.cuttle
+
+sealed trait Metric {
+  def toString: String
+}
+
+case class Gauge(name: String, value: Long, tags: Seq[(String, String)] = Seq.empty) extends Metric
+
+trait MetricProvider {
+  private[cuttle] def getMetrics(jobs: Set[String]): Seq[Metric]
```
guillaumebort: Can't be `private[cuttle]` if it is exposed as part of the Scheduler API.

core/src/main/scala/com/criteo/cuttle/Metrics.scala
```diff
@@ -0,0 +1,22 @@
+package com.criteo.cuttle
```
guillaumebort: Can we scope all this into a `Metrics` object, so in the public API it's all coherent?

core/src/main/scala/com/criteo/cuttle/Metrics.scala
```diff
@@ -0,0 +1,22 @@
...
+  def toString: String
+}
+
+case class Gauge(name: String, value: Long, tags: Seq[(String, String)] = Seq.empty) extends Metric
+
+trait MetricProvider {
+  private[cuttle] def getMetrics(jobs: Set[String]): Seq[Metric]
+}
+
+object Prometheus {
```
guillaumebort: This one should be `private[cuttle]` as it is an implementation detail.

eryshev:
Corrected

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -378,6 +377,32 @@ class Executor[S <: Scheduling] private[cuttle] (
                                          limit: Int): Seq[ExecutionLog] =
     queries.getExecutionLog(queryContexts, jobs, sort, asc, offset, limit).transact(xa).unsafePerformIO
 
+  private[cuttle] def getStats(jobs: Set[String]): Json = {
+    val (running, waiting) = runningExecutionsSizes(jobs)
```
dufrannea: I just wanted to point out that provided figures can be inconsistent because stm accesses are non enclosed in a single 
transaction. Not use this is critical though

timeseries/src/main/scala/com/criteo/cuttle/timeseries/intervals/IntervalMap.scala
```diff
@@ -133,6 +133,10 @@ private[timeseries] object IntervalMap {
         m <- this.intersect(interval).toList
       } yield m): _*))
     }
+
```
dufrannea: You are adding code you do not use, If you really feel this is something that needs to be done I would suggest you could move it to another PR.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/intervals/IntervalMap.scala
```diff
@@ -162,4 +166,6 @@ private[timeseries] sealed trait IntervalMap[A, B] {
   def mapKeys[K: Ordering](f: A => K): IntervalMap[K, B]
   def whenIsDef[C](other: IntervalMap[A, C]): IntervalMap[A, B]
   def whenIsUndef[C](other: IntervalMap[A, C]): IntervalMap[A, B]
+  def head: (Interval[A], B)
```
dufrannea: Same here

timeseries/src/main/scala/com/criteo/cuttle/timeseries/intervals/IntervalMap.scala
eryshev: Cleaned it

eryshev:
@dufrannea All done

PR #177, opened 2017-09-29
Add an init script of precommit hook with scalafmt.

eryshev:


dufrannea:
Cool, could you update the documentation to reflect this ?

PR #179, opened 2017-10-02
Apply Scalafmt to all sources.

eryshev:
Scalafmt for all sources.


vguerci:
@eryshev I guess you have to be careful of the order of merge ;)

eryshev:
Will be done after @guillaumebort doc merge.

PR #182, opened 2017-10-10
Workflow validation on start.

eryshev:
- workflow validation on start
- tests
- git pre-commit correction
Cuttle will System.exit(1) if starting workflow has:
1. some cycles in it
2. (parent, dependency) tuple with parent.startDate before dependency.startDate


timeseries/src/test/scala/com/criteo/cuttle/timeseries/TimeSeriesSpec.scala
```diff
@@ -28,4 +26,73 @@ class TimeSeriesSpec extends FunSuite {
           (job(1), TimeSeriesContext(ts(i), ts(i + 1)))
         })
   }
+
+  test("it should validate empty workflow") {
+    val workflow = Workflow.empty[TimeSeries]
+
+    assume(workflow.isInstanceOf[Workflow[TimeSeries]], "workflow is not timeseries workflow")
```
dufrannea: Why are those type checking preconditions needed ?

dufrannea: I am not sure what the issue is but maybe we should just add type annotation on Workflow.empty ?

guillaumebort:
Why `System.exit`? Because you can start several projects in a singel JVM it's a bit strange to produce a global effect like that from inside the lib. Would be better the throw an exception (or return a Try) and let the launcher fail if it does not handle it.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -261,6 +261,16 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
   }
 
   def start(workflow: Workflow[TimeSeries], executor: Executor[TimeSeries], xa: XA, logger: Logger): Unit = {
+    logger.info("validate workflow before start")
+    TimeSeriesUtils.validate(workflow) match {
+      case Left(errors) =>
+        errors.foreach(logger.error(_))
+        logger.error("exit with error code 1")
+        System.exit(1)
```
guillaumebort: Why System.exit? Because you can start several projects in a singel JVM it's a bit strange to produce a global effect like that from inside the lib. Would be better the throw an exception (or return a Try) and let the launcher fail if it does not handle it.

dufrannea: In that case we should handle the error in CuttleProject if the call to scheduler.start fails, and shutdown the server which is already started up. Currently if scheduler.start throws, the app boots normally.

eryshev:
So for now a start of scheduler throws an exception when workflow is invalid and we don't start the server in this case.

PR #185, opened 2017-10-17
Timeseries: List unresolved parent deps

dufrannea:


guillaumebort:
There is a link on the TODO items, but the link goes nowhere (of course because there is no linked execution yet. Can we link to the period instead? for example /timeseries/executions/hello3/2017-10-16T04Z_2017-10-16T05Z)

core/src/main/javascript/app/components/Badge.js
```diff
@@ -72,7 +72,9 @@ const styles = {
     borderRadius: "2px",
     padding: "0 .5em",
     maxHeight: "18px",
-    textAlign: "center"
+    textAlign: "center",
+    marginRight: "1px",
+    userSelect: "none"
```
guillaumebort: There is a general alignement problem with thses badge. Occuring again in this new screen. Time to fix it? For example with `verticalAlign: "middle"`

core/src/main/javascript/app/pages/TimeSeriesExecutions.js
```diff
@@ -137,17 +141,80 @@ class TimeSeriesExecutions extends React.Component {
...
+      const blockedByExpandedRow = (
+        executions: Array<ExecutionLog>,
+        status: string
+      ) => 
+        _(executions)
+          .filter(e => e.status === status)
+          .sortBy('id')
+          .sortBy('job')
+          .map(e => (
+            <Link href={`/executions/${e.id}`}>
```
guillaumebort: There is a link on the TODO items, but the link goes nowhere (of course because there is no linked execution yet. Can we link to the period instead? for example /timeseries/executions/hello3/2017-10-16T04Z_2017-10-16T05Z)

core/src/main/javascript/app/components/Badge.js
```diff
@@ -72,7 +72,9 @@ const styles = {
     borderRadius: "2px",
     padding: "0 .5em",
     maxHeight: "18px",
-    textAlign: "center"
+    textAlign: "center",
+    marginRight: "1px",
```
guillaumebort: `marginRight: "5px"` would be better.

core/src/main/javascript/app/pages/TimeSeriesExecutions.js
dufrannea: Fixed !

core/src/main/javascript/app/components/Badge.js
dufrannea: Done !

core/src/main/javascript/app/components/Badge.js
dufrannea: I think we should address this in another PR.

PR #186, opened 2017-10-25
Add optional end date to TimeSeries scheduling

haveo:


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -228,13 +228,14 @@ case class TimeSeriesDependency(offset: Duration)
   *                   to a value more than `1` and if possible, the scheduler can trigger [[com.criteo.cuttle.Execution Executions]]
   *                   for more than 1 partition at once.
   */
-case class TimeSeries(calendar: TimeSeriesCalendar, start: Instant, maxPeriods: Int = 1) extends Scheduling {
+case class TimeSeries(calendar: TimeSeriesCalendar, start: Instant, maxPeriods: Int = 1, end: Option[Instant] = None) extends Scheduling {
```
guillaumebort: Why not start, end, maxPeriod then?

PR #187, opened 2017-10-26
allow to pause/resume all executions of one job.

eryshev:
- we can pause/resume filtered set of jobs
- we can pause/resume all executions of one job
- waiting execution queue now completes only jobs awaiting resources
- nuprocess version bump


eryshev:
I added a color for paused jobs in the workflow view

core/src/main/scala/com/criteo/cuttle/platforms/WaitingExecutionQueue.scala
```diff
@@ -53,10 +53,12 @@ trait WaitingExecutionQueue {
     }
     execution
       .onCancel(() => {
-        atomic { implicit txn =>
+        val wasWaiting = atomic { implicit txn =>
```
dufrannea: Could you please comment on this ?

build.sbt
```diff
@@ -187,7 +187,7 @@ lazy val cuttle =
         "org.scala-lang" % "scala-reflect" % scalaVersion.value,
         "org.typelevel" %% "cats" % "0.9.0",
         "codes.reactive" %% "scala-time" % "0.4.1",
-        "com.zaxxer" % "nuprocess" % "1.1.0"
+        "com.zaxxer" % "nuprocess" % "1.1.2"
```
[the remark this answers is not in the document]
eryshev: No need

eryshev:
@dufrannea It's done

core/src/main/scala/com/criteo/cuttle/platforms/WaitingExecutionQueue.scala
```diff
@@ -53,10 +53,13 @@ trait WaitingExecutionQueue {
     }
     execution
       .onCancel(() => {
-        atomic { implicit txn =>
+        // we cancel only waiting executions because the running ones are not in responsibility of the queue.
```
dufrannea: I am not sure this is related to this PR. I'll leave it up to you if you really want to merge this here.

core/src/main/scala/com/criteo/cuttle/App.scala
```diff
@@ -144,7 +143,26 @@ private[cuttle] case class App[S <: Scheduling](project: CuttleProject[S],
   import App._
   import project.{scheduler, workflow}
 
-  private def allJobs = workflow.vertices.map(_.id)
+  private val allJobs = workflow.vertices.map(_.id)
+
+  private def processOnJobIds[T](jobsQueryString: String)(process: Set[String] => T): T =
```
dufrannea: What you did works & does factor some code. 

Though, I think it is preferrable to have something like :
```scala
def parseJobIds(jobsQueryString: String): Set[String] = ???
def getJobsOrNotFound(jobsQueryString: String): Either[Response, Set[Job[S]]] = ???
```
Then use 
```scala
val jobnames = parseJobIds(jobsQueryString) 
```
and
```scala 
case GET at url"/.../?jobs=$jobs" => 
  getJobsOrNotFound.fold(identity, jobNames => {...})
```

This approach :
- creates a simplistic, testable and reusable function getJobsIds (I do not see any good reason why processOnJobIds should require the argument _process_ )
- introduces (minor) complexity where the code is called as opposed to the parsing / lookup functions
- Is, _subjectively_, more readable

I would agree with your approach if you did not have to pass jobs explicitly.

core/src/main/javascript/app/components/SlidePanel.js
```diff
@@ -7,7 +7,8 @@ import classNames from "classnames";
 type Props = {
   classes: any,
   title: string,
-  children: any
+  children: any,
+  open: boolean
```
dufrannea: I am not sure why you extracted the "open" parameter to the props, but I have no objection.

PR #190, opened 2017-11-02
we don't do a double fork in a local platform any more

eryshev:
We don't allow to execute a complex shell scripts in sh interpolation in the sake of a guarantee that forked process will be killed when you want to cancel or pause it.

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
+++ b/core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
@@ -2,24 +2,23 @@ package com.criteo.cuttle.platforms
 
 /** Allow to fork process locally in a managed way.
   *
-  * It provides the `exec` interpolator that you can use to fork shell commands:
+  * It provides the `exec` interpolator that you can use to fork shell scripts:
   *
   * {{{
   *   exec"""hdfs dfs -cp /from /to"""()
   * }}}
   *
   *
-  * The command your provide will be forked into another process.
+  * The script your provide will be forked into another process.
   * Note that if you provide several commands separated by `;' only the first one will be forked
   * and the other ones will be ignored.
   *
   * If you really need to run several commands, you can use one of the followed ways:
   *
   * 1. Chain multiple execs in a for-comprehension. Because of platform mechanism
-  * it will require several platform allocations to run completely. And we can guarantee
-  * that forked process will be killed.
+  * it will require several platform allocations to run completely.
   *
-  * 2. Wrap multiple commands in one `sh -c' _In this case we cannot guarantee that
+  * 2. Wrap it in one `sh -c' _In this case we cannot guarantee that
   * all process spawned inside this fork will be successfully killed when we kill the parent process._
   *
   *
--- a/core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
+++ b/core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
@@ -2,11 +2,17 @@ package com.criteo.cuttle.platforms
 
 /** Allow to fork process locally in a managed way.
   *
-  * It provides the `sh` interpolator that you can use to fork shell scripts:
+  * It provides the `sh` interpolator that you can use to fork shell commands:
   *
   * {{{
   *   sh"""hdfs dfs -cp /from /to""".exec()
   * }}}
+  *
+  * __You command(s) will be executed in a forked `sh -c exec $command` process.
+  * We can provide a guarantee that forked process will be killed when you want
+  * to cancel or pause it only when you put a single command into this interpolation.
+  * If you use multiple commands you should provide a way to kill all potentially forked
+  * children of your process.__
   */
 package object local {
 
--- a/core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
+++ b/core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
@@ -8,46 +8,25 @@ package com.criteo.cuttle.platforms
   *   exec"""hdfs dfs -cp /from /to"""()
   * }}}
   *
-  * Your command(s) will be executed in a forked `sh -c exec $command` process.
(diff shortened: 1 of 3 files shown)


examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
```diff
@@ -45,35 +45,49 @@ object HelloWorld {
...
-          echo "Check my project page at https://github.com/criteo/cuttle"
-          sleep 1
-        """.exec()
+          // declare a bash command to execute. If you want use many commands you can combine them
+          // in a for comprehension.
+          for {
+            _ <- sh"echo Hello for $partitionToCompute".exec()
+            _ <- sh"echo Check my project page at https://github.com/criteo/cuttle".exec()
+            completed <- sh"sleep 1".exec()
+          } yield completed
```
guillaumebort: It would be better to log directly instead of forking a process to log.

```
e.streams.info(s"Hello for $partitionToCompute")
e.streams.info(s"Check my project page at https://github.com/criteo/cuttle")
sh"sleep 1".exec()
```

examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
```diff
@@ -45,35 +45,49 @@ object HelloWorld {
...
-          echo "Hello for ${partitionToCompute}"
-          echo "Check my project page at https://github.com/criteo/cuttle"
-          sleep 1
-        """.exec()
+          // declare a bash command to execute. If you want use many commands you can combine them
+          // in a for comprehension.
+          for {
+            _ <- sh"echo Hello for $partitionToCompute".exec()
+            _ <- sh"echo Check my project page at https://github.com/criteo/cuttle".exec()
+            completed <- sh"sleep 1".exec()
```
guillaumebort: so by the way `sh` name is not relevant anymore.

We could change to something like:

```
exec"sleep 1"()
```

examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
```diff
@@ -45,35 +45,49 @@ object HelloWorld {
...
-          for i in {1..20}; do
-            date
-            sleep 1
-          done
-          echo "Ok"
-        """.exec()
+        val scriptName = s"/tmp/run${e.id}"
+        for {
+          _ <- sh"echo Looping for 20 seconds...".exec()
+          _ <- sh"""
```
guillaumebort: Why doing that instead of an `sh -ce` ? Something like:

```
exec"""sh -ce '
  echo "Looping for 20 seconds..."
  for i in {1..20}; do
    date
    sleep 1
  done
  echo "Ok"
'"""()
```

examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
eryshev: I thought about it but for ui's point of view it looks differently like something that came from framework and not the user code. For example it'll be yellow and not white text. Maybe for the sake of example can we keep it that way ?

examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
eryshev: ```sh"sleep 1".exec()``` becomes ```exec"sleep 1"()```

examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
```diff
@@ -45,35 +45,38 @@ object HelloWorld {
...
-          echo "Hello for ${partitionToCompute}"
-          echo "Check my project page at https://github.com/criteo/cuttle"
-          sleep 1
-        """.exec()
+          // declare a bash command to execute. If you want use many commands you can combine them
+          // in a for comprehension.
+          for {
+            _ <- exec"echo Hello for $partitionToCompute" ()
+            _ <- exec"echo Check my project page at https://github.com/criteo/cuttle" ()
+            completed <- exec"sleep 1" ()
```
guillaumebort: Again I don't think it is a good idea to for a process to log a message. People will probably do the same in their own project by copying the example. Instead use a singe exec""() and replace the echo by pure scala code doing the logging (e.streams.info(...))

examples/src/main/scala/com/criteo/cuttle/examples/HelloWorld.scala
```diff
@@ -97,10 +100,10 @@ object HelloWorld {
     // we need to define the time zone for which _days_ must be considered. The partitions for
     // daily jobs will usually be 24 hours, unless you are choosing a time zone with light saving.
     val world = Job("world", daily(UTC, start), "World") { implicit e =>
-      sh"""
-        echo "World"
-        sleep 6
-      """.exec()
+      for {
```
guillaumebort: Same here.

core/src/main/scala/com/criteo/cuttle/platforms/local/package.scala
```diff
@@ -2,17 +2,54 @@ package com.criteo.cuttle.platforms
...
+  * Example:
+  *
+  * {{{
+  *   exec"""sh -c '
+  *       |    echo 1
+  *       |    echo 2
+  *       |    # we fork a process here that will survive
+  *       |    sh -c "sleep 10000"
+  *       |'"""()
+  * }}}
```
guillaumebort: I think it is better to say something like:


> The command your provide will be forked into another process. Note that if you provide several commands separated by `;' only the first one will be forked and the other ones will be ignored.
> 
> If you really need to run several commands, you can use a shell script (for example using sh -ce '...' etc.).
> 
> Also, note that in case of job cancellation, Cuttle will kill the forked process, but if your process fork other processes (typically the case for a shell script) it is your responsibility to kill them in cascade.

PR #191, opened 2017-11-06
allow to relaunch all stuck jobs

eryshev:
- new route /api/executions/relaunch?jobs=
- new method in Executor `def relaunch` that  is invoked by new route
- code rearrangement in Executor

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Executor.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Executor.scala
@@ -267,7 +267,7 @@ case class Execution[S <: Scheduling](
   /**
     * Run this execution on its job.
     */
-  def run(): Future[Completed] =
+  private[cuttle] def run(): Future[Completed] =
     job.run(this)
 }
 
@@ -629,7 +629,7 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
             execution.streams.error(stacktrace.toString)
             atomic {
               implicit tx =>
-                // retain jobs in a recent failures if a last failure happened in [now - retryStrategy.retryWindow, now]
+                // retain jobs in recent failures if last failure happened in [now - retryStrategy.retryWindow, now]
                 recentFailures.retain {
                   case (_, (retryExecution, failingJob)) =>
                     retryExecution.isDefined || failingJob.isLastFailureAfter(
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -602,10 +605,16 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
...
+    atomic { implicit txn =>
+      execution.startTime = Some(Instant.now())
+      runningState += (execution -> promise.future)
+    }
+
   private def unsafeDoRun(execution: Execution[S], promise: Promise[Completed]): Unit =
     promise.completeWith(
-      execution.job
-        .run(execution)
+      execution
```
dufrannea: Same as above, not sure why it had to change

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -620,6 +629,7 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
             execution.streams.error(stacktrace.toString)
             atomic {
               implicit tx =>
+                // retain jobs in a recent failures if a last failure happened in [now - retryStrategy.retryWindow, now]
```
dufrannea: May I suggest removing both 'a'
// retain jobs in **a** recent failures if **a** last failure happened in [now - retryStrategy.retryWindow, now]

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -262,6 +263,12 @@ case class Execution[S <: Scheduling](
           isWaiting.set(false)
       }
     }
+
+  /**
+    * Run this execution on its job.
+    */
+  def run(): Future[Completed] =
```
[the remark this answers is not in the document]
eryshev: It's  just more convenient to call this instead
```
execution.job.run(execution)
```
But I agree with you that it should be private.

core/src/main/scala/com/criteo/cuttle/Executor.scala
eryshev: You may and I will correct it ty!

PR #198, opened 2017-11-22
Lastruns endpoint

asouletdebrugiere:
Create a new lastrun endpoint that takes a jobId as parameter and which returns
- lastCompleteTime = the end of the first group of successful executions
- lastTime = the end of the last successful execution

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
@@ -7,7 +7,6 @@ import lol.json._
 import io.circe._
 import io.circe.syntax._
 import io.circe.generic.auto._
-
 import scala.util.Try
 import scala.math.Ordering.Implicits._
 import java.time.Instant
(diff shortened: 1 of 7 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -274,7 +274,7 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
 
   val allContexts = Database.sqlGetContextsBetween(None, None)
 
-  private val _state = Ref(Map.empty[TimeSeriesJob, IntervalMap[Instant, JobState]])
+  protected val _state = Ref(Map.empty[TimeSeriesJob, IntervalMap[Instant, JobState]])
```
dufrannea: non necessary, you have already the state in the TimeSeriesApp, use this :

``` scala
      val (state, _) = this.state
```

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -379,6 +381,31 @@ private[timeseries] trait TimeSeriesApp { self: TimeSeriesScheduler =>
       else
         Ok(getCalendar())
 
+    case GET at url"/api/timeseries/lastruns?job=$jobId" =>
+
+      def toFiniteInstant(bound: Bound[Instant]): Instant = bound match {
```
dufrannea: Would be better to handle the fact that a Bound is Finite or not before the serialization phase. The api call will return a 500 with "expected finite instant", which might not be self-explanatory.

PR #202, opened 2017-11-28
allow multiple MySQL endpoints

eryshev:
We allow to configure multiple MySQL endpoints for failover when one of them is unavailable.
When starting a new connection, the driver always tries to connect to the primary host first and, if required, fails over to the secondary hosts on the list sequentially when communication problems are experienced.
https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-config-failover.html

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Database.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Database.scala
@@ -16,7 +16,7 @@ import java.util.concurrent.{Executors, TimeUnit}
 
 import ExecutionStatus._
 
-/** Configuration of MySQL endpoint.
+/** Configuration of JDBC endpoint.
   *
   * @param host JDBC driver host
   * @param port JDBC driver port
@@ -25,6 +25,7 @@ case class DBLocation(host: String, port: Int)
 
 /** Configuration for the MySQL database used by Cuttle.
   *
+  * @param locations sequence of JDBC endpoints
   * @param database JDBC database
   * @param username JDBC username
   * @param password JDBC password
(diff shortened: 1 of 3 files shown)


dufrannea:
I feel it is not really cuttle's concern if you have HA or failover, and thus not its responsibility to build the connectionString. Maybe we should allow to specify a connectionString directly rather than db name / host / port ?

core/src/main/scala/com/criteo/cuttle/Database.scala
```diff
@@ -16,33 +16,42 @@ import java.util.concurrent.{Executors, TimeUnit}
...
+/** Configuration of MySQL endpoint.
   *
   * @param host JDBC driver host
   * @param port JDBC driver port
+  */
+case class DBLocation(host: String, port: Int)
+
+/** Configuration for the MySQL database used by Cuttle.
+  *
   * @param database JDBC database
```
vguerci: locations doc

core/src/main/scala/com/criteo/cuttle/Database.scala
```diff
@@ -178,24 +187,30 @@ private[cuttle] object Database {
     } yield ())
 
   private val connections = collection.concurrent.TrieMap.empty[DatabaseConfig, XA]
-  def connect(c: DatabaseConfig): XA =
+  def connect(dBConfig: DatabaseConfig): XA = {
```
vguerci: OPT: case, should be `dbConfig`

devloop.js
```diff
@@ -17,7 +17,7 @@ let database = runServer({
   name: 'mysqld',
   httpPort: 3388,
   sh: 'java -cp `cat /tmp/classpath_com.criteo.cuttle.localdb` com.criteo.cuttle.localdb.LocalDB',
-}).dependsOn(generateClasspath)
+}).dependsOn(generateClasspath);
```
vguerci: OPT: is that `;` really necessary?

dufrannea: Yes it is ! :)

PR #209, opened 2017-12-13
Prometheus metrics for job's absolute latency

jedirandy:
* In time series scheduler, we expose job latency relative to last successful execution without considering the holes


jedirandy:
Discussed with @vguerci, still need to change the way latency calculated.

PR #210, opened 2017-12-14
Use dedicated ExecutionContexts

Masuzu:
The split is as follows:
- 1 thread pool for Cuttle's HTTP server with low thread limit
- 1 thread pool to run side effects
- 1 thread pool for internal use by side effects (which returns Futures from an execution context)

Other: fixed some typos

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/ExecutionContexts.scala
+++ b/core/src/main/scala/com/criteo/cuttle/ExecutionContexts.scala
@@ -1,0 +1,77 @@
+package com.criteo.cuttle
+
+import java.util.concurrent.Executors
+
+import scala.concurrent.ExecutionContext
+import scala.language.implicitConversions
+import scala.util.Try
+
+object ExecutionContexts {
+
+  sealed trait WrappedExecutionContext {
+    val underlying: ExecutionContext
+  }
+
+  sealed trait Metrics {
+    protected var _threadPoolSize: Long = 0
+    def threadPoolSize(): Long = _threadPoolSize
+  }
+
+  implicit def serverECToEC(ec: ServerExecutionContext): ExecutionContext = ec.underlying
+  implicit def sideEffectECToEC(ec: SideEffectExecutionContext): ExecutionContext = ec.underlying
+  implicit def implicitServerECToEC(implicit ec: ServerExecutionContext): ExecutionContext = ec.underlying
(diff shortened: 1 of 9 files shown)


Masuzu:
I have no clue why the unexisting GitHub co-author 'www' is also referenced.

Masuzu:
Will also add a dedicated execution context for database queries.

Continuous integration is so bad Q_Q

Masuzu:
The main issue with mixing together different implicit typed ExecutionContexts with calls to methods taking an implicit untyped ExecutionContext (Future creation for instance) as argument is that you will run into ambiguous conversions from those typed executions to the untyped ExecutionContext: for instance, importing `ExecutionContexts.Implicits._` in Executor.scala will result in issues such as this one:

```
Error:(441, 63) ambiguous implicit values:
both method implicitServerECToEC in object ExecutionContexts of type (implicit ec: com.criteo.cuttle.ExecutionContexts.ServerExecutionContext)scala.concurrent.ExecutionContext
and method implicitSideEffectECToEC in object ExecutionContexts of type (implicit ec: com.criteo.cuttle.ExecutionContexts.SideEffectExecutionContext)scala.concurrent.ExecutionContext
match expected type scala.concurrent.ExecutionContext
     .firstCompletedOf(List(Timeout(timeout), Future.sequence(runningFutures)))
```

loicknuchel:
Yes, and that's also a nice part ;)
You should mainly have implicit parameters with typed execution contexts and only import them in a few narrow scopes.

core/src/main/scala/com/criteo/cuttle/ExecutionContexts.scala
```diff
@@ -0,0 +1,77 @@
...
+import scala.language.implicitConversions
+import scala.util.Try
+
+object ExecutionContexts {
+
+  sealed trait WrappedExecutionContext {
+    val underlying: ExecutionContext
+  }
+
+  sealed trait Metrics {
```
jedirandy: what is this for? to count the total threads created during the lifetime? A little comment can be helpful.

Masuzu: Yes it is, to be renamed as threadPoolSize then

Masuzu:
@jedirandy Updated. Let me know if it is still unclear.

PR #211, opened 2017-12-14
Prometheus metric for absolute latency of time series job

jedirandy:
* Expose absolute latency in seconds of each time series job as a Prometheus metric


jedirandy:
Comments of Prometheus metrics updated

PR #212, opened 2017-12-15
remove logging.properties and pass projectName instead of CuttleProject to Executor

eryshev:
- logging.properties isn't used
- CuttleProject was passed down to Executor and injected to each Execution. So we had execution.cuttleProject.scheduler api available to end user. I don't think that is a good idea so I removed cuttleProject and passed projectName instead.


eryshev:
Updated it.

PR #215, opened 2017-12-27
remove debug output from local platform

eryshev:
We need to do this because forked command can contain credentials and we don't want to show it to users.


vguerci:
Also, I think we want to show the executed command. For not showing passwords, there are other possibilities such as masking *** or communicating them through env vars

eryshev:
Okey I will make change in our internal implementation.

PR #218, opened 2018-01-15
Add API endpoint to ignore failing executions and mark them as successful

Masuzu:


core/src/main/scala/com/criteo/cuttle/App.scala
```diff
@@ -364,6 +364,16 @@ private[cuttle] case class App[S <: Scheduling](project: CuttleProject[S], execu
       IO.pure(Ok)
     }
 
+    case POST at url"/api/executions/forceSuccess?jobs=$jobs" => { implicit user =>
```
dufrannea: I do not think this is the endpoint we need. I was thinking more of an endpoint to force success on a single execution. Here you are forcing success on a potentially big number of executions which, I think, should be avoided.

Masuzu: I will make the change. That makes sense.

Masuzu:
Updated and tested locally.

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -629,7 +661,7 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
     promise.completeWith(
       (Try(execution.run()) match {
         case Success(f) => f
-        case Failure(e) => Future.failed(e)
+        case Failure(e) => if (execution.shouldForceSuccess) Future { Completed } else Future.failed(e)
```
eryshev: Use Future.successful

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -150,10 +150,17 @@ case class Execution[S <: Scheduling](
...
+  /**
+    * An execution with forcedSuccess set to true will have its side effect return a successful Future instance even if the
+    * user code raised an exception or returned a failed Future instance.
+    */
+  private val forcedSuccess = Ref(false)
 
   /** Returns `true` if this [[Execution]] has been cancelled. */
   def isCancelled = cancelled.single()
 
+  def shouldForceSuccess = forcedSuccess.single()
```
eryshev: Is it useful ?

Masuzu: It is used at line 664 and is used to separate the atomic boolean `forcedSuccess` from the actual "view" of that object, in the same way `isCancelled` was used.

core/src/main/scala/com/criteo/cuttle/Executor.scala
Masuzu: Will do

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -203,6 +210,23 @@ case class Execution[S <: Scheduling](
...
+  def forceSuccess()(implicit user: User): Boolean = {
+    val shouldForceSuccess = atomic { implicit txn =>
+      if (forcedSuccess()) {
+        false
+      } else {
+        forcedSuccess() = true
+        true
+      }
+    }
+    if (shouldForceSuccess) {
```
eryshev: It isn't necessary, just check focedSucces

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -203,6 +210,23 @@ case class Execution[S <: Scheduling](
     hasBeenCancelled
   }
 
+  def forceSuccess()(implicit user: User): Boolean = {
```
eryshev: You don't use a return type so I think it can be Unit

core/src/main/scala/com/criteo/cuttle/App.scala
```diff
@@ -331,6 +331,11 @@ private[cuttle] case class App[S <: Scheduling](project: CuttleProject[S], execu
       Ok
     }
 
+    case POST at url"/api/executions/$id/forceSuccess" => { implicit user =>
```
eryshev: OPT: Personally I prefer force/success :)

Masuzu: Done

core/src/main/scala/com/criteo/cuttle/Executor.scala
Masuzu: Done

core/src/main/scala/com/criteo/cuttle/Executor.scala
Masuzu: Removed it

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -203,6 +210,20 @@ case class Execution[S <: Scheduling](
     hasBeenCancelled
   }
 
+  def forceSuccess()(implicit user: User): Unit = {
+    atomic { implicit txn =>
+      if (forcedSuccess()) {
+        false
```
eryshev: Could we delete it as it isn't used? Just something like this
```
if (!forcedSuccess)
forcedSuccess() = true
print
```

core/src/main/scala/com/criteo/cuttle/Executor.scala
eryshev: I suggest that we delete it and use `forcedSuccess.single()` at line 664

core/src/main/scala/com/criteo/cuttle/Executor.scala
Masuzu: Yes sure

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -203,6 +210,16 @@ case class Execution[S <: Scheduling](
     hasBeenCancelled
   }
 
+  def forceSuccess()(implicit user: User): Unit =
+    atomic { implicit txn =>
+      if (!forcedSuccess()) {
+        forcedSuccess() = true
+        streams.debug(
```
Masuzu: No

core/src/main/scala/com/criteo/cuttle/Executor.scala
Masuzu: Nuke

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -629,7 +651,7 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
     promise.completeWith(
       (Try(execution.run()) match {
         case Success(f) => f
-        case Failure(e) => Future.failed(e)
+        case Failure(e) => if (execution.forcedSuccess.single()) Future.successful(Completed) else Future.failed(e)
```
loicknuchel: so when we force an execution at success, it will stay in running / paused or other until it finised ?

Masuzu: Yes that's right

PR #222, opened 2018-01-25
Add Weekly calendar

Lordshinjo:


eryshev:
Don't we have to adjust our UI for it?
You can check this file for example https://github.com/criteo/cuttle/blob/418bdbfac3a38315371538f2092f5c2318563063/core/src/main/javascript/datamodel.js#L54

Lordshinjo:
I saw that one, but I got a bit confused, because Monthly isn't there either.
I guess both should be there then.

eryshev:
UI lgtm

PR #223, opened 2018-01-30
Separate low and high bounds in TimeSeriesDependency

haveo:
This allows to depend on multiple periods through one dependency. The
meaning of the offset has also been reversed. An offset is now applied
to the desired period to get the period the dependency has to satisfy.

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
@@ -9,8 +9,10 @@ import java.util.UUID
 import scala.concurrent._
 import scala.concurrent.duration.{Duration => ScalaDuration}
 import scala.concurrent.stm._
+import scala.math.Ordering.Implicits._
 
-import cats.Eq
+import cats._
+import cats.implicits._
 import cats.effect.IO
 import cats.mtl.implicits._
 import doobie._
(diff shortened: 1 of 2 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -527,20 +535,32 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
     val parentsMap = workflow.edges.groupBy { case (child, _, _)   => child }
     val childrenMap = workflow.edges.groupBy { case (_, parent, _) => parent }
 
+    def reverseDescr(dep: TimeSeriesDependency) =
+      TimeSeriesDependency(dep.offsetHigh.negated, dep.offsetLow.negated)
+    def applyDep(dep: TimeSeriesDependency) = Function.unlift { (i: Interval[Instant]) =>
```
loicknuchel: it can be better to have the function return type... Not easy to understand without IDE...

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -527,20 +535,32 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
+      val low = i.lo.map(_.plus(dep.offsetLow))
+      val high = i.hi.map(_.plus(dep.offsetHigh))
+      if (low >= high) None else Some(Interval(low, high))
```
loicknuchel: so we just ignore dependencies with low >= high ?
Maybe this check could be done in a test and at the beginning when the workflow is initialized to fail the startup ?

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
haveo: sure

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
haveo: No, that's by design. If you have `job1 dependsOn ((job2, TimeSeriesDependency(1.hours, 0.hours)))` that means job1 needs the current hour of job2 plus the previous one. If you have period `[00:00; 02:00[` for job2 the scheduler will compute that you can execute job1 for the period `[01:00; 02:00[` (it just adds the bounds in the dependency descriptor). If you have period `[00:00; 01:00[` for job2 it will compute that you can execute job1 for period `[01:00; 01:00[` so by discarding this you get the wanted result: you can't execute job1 as you only have one hour available for job2.

PR #225, opened 2018-02-05
Create server outside of cuttle

loicknuchel:
This change allow a more flexible way to start Cuttle. You can now get only needed routes and bootstrap the server externally (to add some routes or custom behaviour)

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/CuttleProject.scala
+++ b/core/src/main/scala/com/criteo/cuttle/CuttleProject.scala
@@ -49,6 +49,31 @@ class CuttleProject[S <: Scheduling] private[cuttle] (
 
     logger.info(s"Listening on http://localhost:$httpPort")
   }
+
+  /**
+    * Connect to database and build routes. It allows you to start externally a server and decide when to start the scheduling.
+    *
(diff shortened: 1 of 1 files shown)


loicknuchel:
For context, the Dataflow review using this is here: http://review.criteois.lan/#/c/317658/

core/src/main/scala/com/criteo/cuttle/CuttleProject.scala
```diff
@@ -49,6 +49,31 @@ class CuttleProject[S <: Scheduling] private[cuttle] (
...
+    */
+  def build(
+    platforms: Seq[ExecutionPlatform] = CuttleProject.defaultPlatforms,
+    databaseConfig: DatabaseConfig = DatabaseConfig.fromEnv,
+    retryStrategy: RetryStrategy = RetryStrategy.ExponentialBackoffRetryStrategy
+  ): (Service, () => Unit) = {
+    val xa = Database.connect(databaseConfig)
+    val executor = new Executor[S](platforms, xa, logger, name)(retryStrategy)
+
+    val startScheduler = () => {
```
guillaumebort: No need to expose that again. It is already available to user code.

PR #234, opened 2018-02-20
Refactor graph validation to reduce complexity from O(VE) to O(V+E)

Masuzu:
V being the number of jobs and E the number of dependencies


core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -56,8 +56,14 @@ trait Workflow[S <: Scheduling] {
     }
   }
 
-  private[cuttle] lazy val roots = vertices.filter(v => edges.forall { case (v1, _, _)  => v1 != v })
-  private[cuttle] lazy val leaves = vertices.filter(v => edges.forall { case (_, v2, _) => v2 != v })
+  private[cuttle] lazy val roots = {
```
eryshev: format trailling lines

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -730,32 +730,47 @@ private[timeseries] object TimeSeriesUtils {
     */
   def validate(workflow: Workflow[TimeSeries]): Either[List[String], Unit] = {
```
dufrannea: Would it be possible to expose this on Workflow ? use case is to validate a workflow externally before sending it to cuttle.

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: I do not see any trailing line issue except no empty line between the definitions of roots and leaves?

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
Masuzu: Yes sure will do

core/src/main/scala/com/criteo/cuttle/Workflow.scala
eryshev: that is it

Masuzu:
@dufrannea I'll add a second commit with Strongly connected components detection to list all the jobs involved in cycles then. It has the same linear complexity as the original implementation.

PR #235, opened 2018-02-20
Add max parallel executions configuration parameter for jobs

Masuzu:


haveo:
I don't think we should put more things like this in the Executor but instead add them as platforms. IMHO, `synchronize` shouldn't be there either and even the pause could be factored out. Basically, waiting on things is the job of execution platforms which provide a more modular design.

@guillaumebort thoughts?

Masuzu:
As discussed, will drop this change.

The use case would be to have some kind of:

```
semaphore_with_(user_configurable_threshold).acquire {
  execute workflow subset
}
```

with a `semaphore_with_(user_configurable_threshold)` defined in a similar way as the existing `synchronized`, instead of creating a different platform for each workflow subset/job which needs a control over the max parallelism..

PR #236, opened 2018-02-21
Add workflow helper method to identify the cycles in the workflow in linear time

Masuzu:


core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,145 @@ trait Workflow[S <: Scheduling] {
...
+        if (!child.visited) {
+          // (node -> child) is a tree edge
+          dfs(child)
+          node.earliestAncestor = Math.min(node.earliestAncestor, child.earliestAncestor)
+        } else {
+          if (nodesInStack.contains(child)) {
+            // (node -> child) is a back edge
+            node.earliestAncestor = Math.min(node.earliestAncestor, child.discoveryTime)
           }
+          // Otherwise (node -> child) is a cross edge and 'child' was already added to a SCC
```
Masuzu: Actually it is not necessarily a cross edge. It could be a tree edge or a forward edge to a node already added to a SCC .

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
+  }
+
+  /**
+    * @param edgeVisitor method called for each edge (parent node, child node) visited by Kahn's algorithm
+    * @return the jobs in the workflow sorted in topological order, None if the graph contains a cycle
+    * */
+  def topologicalSort(edgeVisitor: (Job[S], Job[S]) => Unit = (_, _) => ()): Option[List[Job[S]]] = {
+    val topologicalOrder = mutable.ListBuffer.empty[Job[S]]
+
+    case class NodeWithMetadata(node: Job[S], children: mutable.Set[NodeWithMetadata] = mutable.Set.empty) {
```
loicknuchel: why not JobWithMetadata or VerticeWithMetadata ? Seems good to keep the same vocabulary everywhere...

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
+      def isOrphanNode: Boolean = numIncomingEdges == 0
+
+      // Override hash computation to prevent stack overflow when trying to compute the hash of a node using the
+      // default hash method on workflows with cycles
+      override def hashCode(): Int = node.hashCode()
     }
 
-    require(edges.isEmpty, "Workflow has at least one cycle")
-    result.toList
+    val nodesWithMetadata = vertices.map(job => NodeWithMetadata(job))
```
loicknuchel: I think `val nodesWithMetadata = vertices.map(job => (job, NodeWithMetadata(job))).toMap` is more readable...

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
+
+    stronglyConnectedComponents.filter(_.size >= 2).toList
+  }
+
+  /**
+    * @param edgeVisitor method called for each edge (parent node, child node) visited by Kahn's algorithm
+    * @return the jobs in the workflow sorted in topological order, None if the graph contains a cycle
+    * */
+  def topologicalSort(edgeVisitor: (Job[S], Job[S]) => Unit = (_, _) => ()): Option[List[Job[S]]] = {
+    val topologicalOrder = mutable.ListBuffer.empty[Job[S]]
```
loicknuchel: maybe move topologicalOrder declaration next to rootsToVisit one (as they are used together)

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -90,6 +202,18 @@ trait Workflow[S <: Scheduling] {
...
+  case class DependencyBuilder(workflow: Workflow[S], parentJob: Job[S])(implicit dependencyDescriptor: S#DependencyDescriptor) {
+    def to(childJob: Job[S]): Workflow[S] = {
+      new Workflow[S] {
+        val vertices = workflow.vertices ++ Set(childJob, parentJob)
+        val edges = workflow.edges + ((childJob, parentJob, dependencyDescriptor))
+      }
+    }
+  }
+
+  def withDependencyFrom(parentJob: Job[S])(implicit dependencyDescriptor: S#DependencyDescriptor): DependencyBuilder =
```
loicknuchel: not sure it's worth it to add new API method just for tests. I think it's an important design decision to discuss before doing it

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
-      result.append(root)
-      val edgesToRemove = edges.filter(_._2 == root)
-
-      edgesToRemove.foreach {
-        case edge @ (child, _, _) =>
-          edges.remove(edge)
-          if (!edges.exists(_._1 == child)) {
-            orderedRoots.add(child)
-          }
+  def roots: Set[Job[S]] = {
```
loicknuchel: not sure we want to expose theses methods in API...

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
+    }
+
+    stronglyConnectedComponents.filter(_.size >= 2).toList
+  }
+
+  /**
+    * @param edgeVisitor method called for each edge (parent node, child node) visited by Kahn's algorithm
+    * @return the jobs in the workflow sorted in topological order, None if the graph contains a cycle
+    * */
+  def topologicalSort(edgeVisitor: (Job[S], Job[S]) => Unit = (_, _) => ()): Option[List[Job[S]]] = {
```
loicknuchel: could be nice to have this as a pure & generic function (in a utils package for example)

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
+    case None => throw new IllegalArgumentException("Workflow has at least one cycle")
+  }
+
+  /**
+    * @return the list of strongly connected components (SCCs) in the workflow with more than 1 node, which
+    * are the cycles in the graph
+    * @note Tarjan algorithm
+    */
+  def findCycles(): List[List[Job[S]]] = {
+    case class NodeWithMetadata(node: Job[S], children: mutable.Set[NodeWithMetadata] = mutable.Set.empty) {
```
loicknuchel: why not JobWithMetadata or VerticeWithMetadata ? Seems good to keep the same vocabulary everywhere...

core/src/main/scala/com/criteo/cuttle/Workflow.scala
```diff
@@ -16,30 +17,144 @@ trait Workflow[S <: Scheduling] {
...
+    case Some(sortedJobs) => sortedJobs
+    case None => throw new IllegalArgumentException("Workflow has at least one cycle")
+  }
+
+  /**
+    * @return the list of strongly connected components (SCCs) in the workflow with more than 1 node, which
+    * are the cycles in the graph
+    * @note Tarjan algorithm
+    */
+  def findCycles(): List[List[Job[S]]] = {
```
loicknuchel: could be nice to implement it as a generic & pure function and call it here.
Not sure it we want to expose this in the API...

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Will do

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Sure, will do

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Agreed, will do

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Will do

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Totally, will do!

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Agreed. Will remove it.

core/src/main/scala/com/criteo/cuttle/Workflow.scala
Masuzu: Will do.

Masuzu:
Superseded by https://github.com/criteo/cuttle/pull/244

PR #239, opened 2018-02-26
Add helper to lock an async code block execution behind a semaphore

Masuzu:


Masuzu:
I'll add a check to perform that check at runtime

PR #244, opened 2018-02-28
Refactor graph validation to reduce complexity from O(VE) to O(V+E)

Masuzu:
V being the number of jobs and E the number of dependencies

Supersedes https://github.com/criteo/cuttle/pull/236 taking into @loicknuchel comments and drops the cycle identification for this change.

See `TimeSeriesUtils.validate()`

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/graph/mutable/Graph.scala
+++ b/core/src/main/scala/com/criteo/cuttle/graph/mutable/Graph.scala
@@ -1,0 +1,38 @@
+package com.criteo.cuttle.graph.mutable
+
+import scala.collection.mutable
+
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/criteo/cuttle/graph/mutable/Graph.scala
```diff
@@ -0,0 +1,38 @@
+package com.criteo.cuttle.graph.mutable
+
+import scala.collection.mutable
+
+case class Node[T](data: T, children: mutable.Set[Node[T]] = mutable.Set.empty[Node[T]]) {
```
dufrannea: private[cuttle]

PR #253, opened 2018-03-13
timeseries: add a class to persist start dates in mysql

haveo:


haveo:
There's probably a conflict with the locking transactor. I initially thought of using the transactor created by Cuttle. However the transactor is built in CuttleProject which is problematic since we need the workflow to instantiate CuttleProject but we want the transactor to build the workflow. I don't see any way to fix this without breaking the API.

Thoughts?

guillaumebort:
I'm not sure to understand why we want to persist start dates this way in the database. I understand the issue of having startDate configurable thus not in the code, but what this code is intended to do?

haveo:
Well it's a step towards having startDate configuration fully supported. You can use it to enable a "startDate is the first deploy date" use case or you could add HTTP routes and do it all manually.

guillaumebort:
Using the same database does not necessary mean using the same connection pool. Perhaps we don't need to share the same connection pool. Otherwise we will effectively have some loop in the dependency between the user code and cuttle.

I see 2 possibilities:

- Changing the API so creating the connection pool is the responsability of the user code. Then the user can give it to cutlle to init the storage and also use it the way it want.
- Letting cuttle manage its own connection pool, and then let the user code create its own connection pool if it want.

haveo:
Ended up going the second route. Cuttle uses its locked transactor and the client can create its own unlocked transactor if it wants to.

core/src/main/scala/com/criteo/cuttle/Database.scala
```diff
@@ -119,6 +114,11 @@ private[cuttle] object Database {
 
     // Try to insert our lock at bootstrap
     (for {
+      _ <- sql"""CREATE TABLE IF NOT EXISTS locks (
```
guillaumebort: Is this related to your change? Or just some additional refactoring?

Because I don't understand how this impact the feature you expose. Here.

haveo: Yep there are some kind of unrelated changes here, I added them in the commit message. Basically I'm locking the database while doing schema migration so we have to move the lock table creation out of the schema migration.

PR #261, opened 2018-03-23
introducing more robust statistics to empower users

loicknuchel:


eryshev:
Where are unit tests ?

PR #262, opened 2018-03-29
Resolve backfills on valid subperiods

Masuzu:
Automatically resolve valid backfills on subperiods instead of requesting the backfill to be valid on the whole input period.
Assume a backfill is requested over the period t1, t2 for the jobs = {job1, job2}
For illustrative purposes, let the past execution state be as follows:
```
| time | t1 | t2 | t3 | t4 | t5 |
| job1 | OK | OK | KO | OK | OK |
| job2 | OK | OK | KO | KO | OK |
````

Then the backfill will be lauched on {t1, t2} for jobs {job1, job2}, on {t4} for {job1} and on {t5} for {job1, job2}

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
@@ -6,15 +6,17 @@ import java.time.temporal.ChronoUnit._
 import java.time.temporal.{ChronoUnit, TemporalAdjusters}
 import java.util.UUID
 
+import scala.collection.mutable
 import scala.concurrent._
 import scala.concurrent.duration.{Duration => ScalaDuration}
 import scala.concurrent.stm._
 import scala.math.Ordering.Implicits._
 
 import cats._
-import cats.implicits._
 import cats.effect.IO
+import cats.implicits._
 import cats.mtl.implicits._
+import de.sciss.fingertree.RangedSeq
 import doobie._
 import doobie.implicits._
 import io.circe._
@@ -323,7 +325,7 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
     (_state(), _backfills())
   }
 
-  private case class BackfillError(job: TimeSeriesJob, msg: String) {
+  private[timeseries] case class BackfillError(job: TimeSeriesJob, msg: String) {
     override def toString: String =
       s"""
          |Error during backfill creation.
@@ -336,7 +338,13 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
   private def assert(condition: Boolean, msg: String)(implicit job: TimeSeriesJob) =
     (if (condition) Right(true) else Left(BackfillError(job, msg))).right
 
-  private def validateAndUpdateState(backfill: Backfill) = atomic { implicit txn =>
+  /**
+    * @return the list of errors for the input backfill configuration, if any
+    */
+  private[timeseries] def validateBackfill(
+    backfill: Backfill,
+    states: Map[TimeSeriesUtils.TimeSeriesJob, IntervalMap[Instant, JobState]]
+  ): Set[BackfillError] = {
     val start = backfill.start
     val end = backfill.end
     val validationByJob = backfill.jobs.map { implicit job =>
@@ -345,7 +353,7 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
         validIn <- {
           val interval = Interval(start, end)
           // collect all periods that are intersecting with [start, end]
-          val interval2State = _state().apply(job).intersect(interval).toList
+          val interval2State = states.apply(job).intersect(interval).toList
           // in a Done state, and correct periodicity
           Right(
             interval2State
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by dufrannea** [comment]
The issue with this approach is that you do not guarantee that failing jobs that are running at the moment of the backfill (so that are not backfilled according to you definition) will run in their latest version (I particularly think of remote jobs that can reattach on restart), which is I believe what is intended when you backfill a period. One solution would be to cancel all running jobs for the given period.

**Comment by Masuzu** [comment]
Ah, ok, I get the point, thank you. I will go with the cancel solution.
There will be a race condition if cancelling the execution fails and the execution which is still running succeeds after the actual backfill completes. I will see what I can do, probably waiting for all the executions still running on the period to complete before launching the backfill.

**Comment by Masuzu** [comment]
Updated the review to take into account cancellation of conflicting running executions.

File: timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala [hunk 1]
```diff
@@ -370,44 +374,150 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
+    val candidateBackfillsByPeriod: List[((Instant, Instant), TimeSeriesJob)] = jobs.flatMap { job =>
+      // collect all periods that are intersecting with [start, end]
+      val interval2State = states(job).intersect(interval).toList
+      // in a Done state, and correct periodicity
+      interval2State.collect { case (Interval(Finite(lo), Finite(hi)), Done) =>
+        (lo, hi) -> job
+      }
+    }.toList
+
+    val jobsToBackFillByPeriod = candidateBackfillsByPeriod.groupBy { case (interval, job) => interval }
```
dufrannea: This wont be able to group correctly. Say for example job1 is backfillable from 1 to 3 & jobs depending of job1 from 1 to 2, they will be created as separate backfills.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -370,44 +374,150 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
+      .map { case (interval, groups) =>
+        interval -> groups.map { case (interval, job) => job }.toSet
+      }
+
+    val backfills = mutable.ArrayBuffer.empty[Backfill]
```
dufrannea: not sure why we need a mutable ArrayBuffer here

Masuzu:
TODO:
Backfills have to be lauched on the smallest intervals of jobs with the state "Done", and each of these intervals, the backfill to create must include all the jobs in that interval. This is required to avoid race conditions with groups of jobs with dependencies to backfill, or else child dependencies may run first before their parents, which is not the expected behaviour.

timeseries/src/test/scala/com/criteo/cuttle/timeseries/TimeSeriesSpec.scala
```diff
@@ -111,4 +111,55 @@ class TimeSeriesSpec extends FunSuite with TestScheduling {
...
+    val job1 = Job("job1", hourly(date"2117-03-25T02:00:00Z"))(completed)
+    val job2 = Job("job2", hourly(date"2117-03-24T02:00:00Z"))(completed)
+
+    // hours 1 2 3 4 5 6 7
+    // job1  VVVVVV--VVVVV
+    // job2  --VVVV----VVV
+    // V means that the job was done on that period. Periods are indivisible.
+    // Backfill requested from hour 2 to 7
+    // Backfills on periods overlapping valid calendar periods are automatically resolved when the actual backfill is run.
+    val jobStates = Map[TimeSeriesUtils.TimeSeriesJob, IntervalMap[Instant, JobState]](
```
dufrannea: This setup shoud create 3 backfills : 
- 2-4 (job1, job2)
- 5-6 job1
- 6-7 (job1, job2)

Masuzu: Indeed, this will be addressed.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -336,7 +337,10 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
   private def assert(condition: Boolean, msg: String)(implicit job: TimeSeriesJob) =
     (if (condition) Right(true) else Left(BackfillError(job, msg))).right
 
-  private def validateAndUpdateState(backfill: Backfill) = atomic { implicit txn =>
+  private[timeseries] def validateBackfill(
+    backfill: Backfill,
+    states: Map[TimeSeriesUtils.TimeSeriesJob, IntervalMap[Instant, JobState]]
+  ): (Boolean, Set[Either[BackfillError, Boolean]]) = {
```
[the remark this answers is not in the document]
Masuzu: Thank you, yes I will get rid of this nonsensical return type!

Masuzu:
All the remarks should have been taken into account!
The IDE killing my VM took nearly more time than making the actual changes...

timeseries/src/test/scala/com/criteo/cuttle/timeseries/TimeSeriesSpec.scala
Masuzu: Done

Masuzu:
Addressed the issue with side effect in a STM atomic block

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -369,45 +377,179 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
-    )
+    val errors = mutable.ArrayBuffer.empty[String]
+    val (validBackfills, futureCancelledExecutions): (List[Backfill], Future[List[Completed]]) = atomic { implicit txn =>
+      val backfills = createBackfills(name, description, jobs, _state(), start, end, priority)
+
+      val validBackfills = backfills.flatMap { newBackfill =>
+        validateBackfill(newBackfill, _state()).toList match {
+          case Nil =>  Some(newBackfill)
+          case (validationErrors) =>
+            errors += validationErrors.mkString("\n")
```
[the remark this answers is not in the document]
Masuzu: I did the update if I am not mistaken?
That was the old code. Now errors is instantiate inside the atomic STM block.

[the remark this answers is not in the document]
dufrannea: Sorry apparently I did not publish this comment during my latest review and it somehow got added now with my approval :) sorry about that you can disregard it

PR #263, opened 2018-04-03
Fix graph node display issue

jedirandy:
- Fix the issue when some job names cannot be correctly displayed

### DIFF
--- a/core/src/main/javascript/graph/dagger/render/nodesAndEdges.js
+++ b/core/src/main/javascript/graph/dagger/render/nodesAndEdges.js
@@ -1,7 +1,7 @@
 import { nodeKind, edgeKind } from "../layout/symbolic/annotatedGraph";
 import * as d3 from "d3";
 import { textWrap } from "d3plus-text";
-import { forEach, identity } from "lodash";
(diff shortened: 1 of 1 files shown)


core/src/main/javascript/graph/dagger/render/nodesAndEdges.js
```diff
@@ -277,6 +277,13 @@ const tagBulletVerticalOffset = ({ height }) => {
     startOffset + tagIndex * (bulletSize + spaceBetweenBullets);
 };
 
+// join words by whitespace, unless it's delimited by -_
+const joinWords = (words: Array<string>): string =>
+  reduce(
+    words,
+    (acc, word) => (/[-_]$/.test(acc) ? acc + word : acc + " " + word)
```
Masuzu: Shouldn't the regex test `word` instead of the accumulator?

jedirandy: ["this-","is-","it"] consider this sentence, you want to check the suffix of the accumulated result 
 then append each word (the first word being the initial acc)

PR #266, opened 2018-04-05
Add GET /version endpoint to rest API

Masuzu:


vguerci:
Live coding in gh editor ftw ;p

PR #277, opened 2018-04-17
Custom scheduling

guillaumebort:
Replace the core `Workflow` concept with a more generic `Workload` one without any notion of dependencies.

----

The idea is that the current `Workflow` is very tied to the timeseries scheduler. It can be a useless concept for simpler scheduler without any need to express dependencies. At the same time it is not powerful enough for more sophisticated scheduler that could manage more dynamic workflow.

The current `Worklow` type is moved to the timeseries package. In the core model we introduce a new `Workload` type that is basically a set of `Jobs' to execution.

We also introduce a `HelloCustomScheduling` example demonstrating how one can use Cuttle as a platform to execute jobs with an adhoc scheduling mechanism.

----

On the UI side it is more complicated as a lot of timeseries screens are hard-coded in the core module. This change ensure that the UI is still compatible even if another scheduler is used, but some screens are not available.

More work is needed in the future to allow custom schedulers to plug their own UI in.


Masuzu:
I am afraid that a manual rebase will have to be made.

PR #286, opened 2018-05-16
Update Flow version

haveo:
Also, fix a display bug for stuck executions


core/src/main/javascript/app/components/PopoverMenu.js
```diff
@@ -11,17 +11,14 @@ type Props = {
   classes: any,
   className: string,
   envCritical: boolean,
-  items: Array<Node>
+  items: Array<any>
```
eryshev: Why can't we put a proper type here?

PR #289, opened 2018-06-28
[EXPERIMENTAL] slff4J & Log4J redirection

BenoitHanotte:
Slf4j and log4j are the standard for Java logging (eg: used by hadoop, hive, spark, ...). 
However when used from within Cuttle, these logs are not routed through the `ExecutionStreams`and thus not available in Cuttle's execution view, making it very hard to debug individual executions as these logs often end-up all mixed together in `stdout`.

This PR is a proof of concept on how we can wrap the `SideEffectExecutionContext` to register in a Map, each running thread (~future) of a side effect and the associated ExecutionStreams.
When a slf4j log event is then emitted by one of these thread, it is intecepted by a custom log4j appender, `ExecutionStreamsAppender`, which will then get the `ExecutionStreams` associated with the Thread which emitted this event and route the message through the streams. The event messages then become available in the Cuttle UI.

As show in `HelloWorldSLF4J`, this also works when multiple futures are running in parallel for a single side-effect.

This commit is not ready to be merged as it is, but is rather intended at being a proof of concept to allow further discussion on the subject.

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Executor.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Executor.scala
@@ -2,6 +2,7 @@ package com.criteo.cuttle
 
 import java.io.{PrintWriter, StringWriter}
 import java.time.{Duration, Instant, ZoneId}
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.{Timer, TimerTask}
 
@@ -349,6 +350,15 @@ private[cuttle] object ExecutionPlatform {
   implicit def fromExecution(implicit e: Execution[_]): Seq[ExecutionPlatform] = e.platforms
   def lookup[E: ClassTag](implicit platforms: Seq[ExecutionPlatform]): Option[E] =
     platforms.find(classTag[E].runtimeClass.isInstance).map(_.asInstanceOf[E])
+}
+
+private[cuttle] object Executor {
+
+  // we save a mapping of ThreadName -> ExecutionStreams to be able to redirect logs comming
+  // form different SideEffect (Futures) to the corresponding ExecutionStreams
+  private val threadNamesToStreams = new ConcurrentHashMap[String, ExecutionStreams]
+
+  def getStreams(threadName: String): Option[ExecutionStreams] = Option(threadNamesToStreams.get(threadName))
 }
 
 /** An [[Executor]] is responsible to actually execute the [[SideEffect]] functions for the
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -827,16 +847,29 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
...
+                  val streams = new ExecutionStreams {
+                    def writeln(str: CharSequence) = ExecutionStreams.writeln(nextExecutionId, str)
+                  }
+
+                  // wrap the execution context so that we can register the name of the thread of each
+                  // runnable (and thus future) that will be run by the side effect.
+                  val sideEffectExecutionContext = SideEffectExecutionContext.wrap(runnable => new Runnable {
+                      override def run(): Unit = {
+                        val tName = Thread.currentThread().getName
+                        Executor.threadNamesToStreams.put(tName, streams)
```
guillaumebort: Thread names are not guaranteed to be unique. Why not put a reference to the `Thread` instead?

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -351,6 +353,24 @@ private[cuttle] object ExecutionPlatform {
     platforms.find(classTag[E].runtimeClass.isInstance).map(_.asInstanceOf[E])
 }
 
+private[cuttle] object Executor {
+
+  // CAUTION: we use a Weak value hash map provided by guava for 2 reasons:
```
guillaumebort: I don't really get it. As we control the code accessing this Map I don't see any case where we would loose a reference to the `ExecutionStream`. Plus it would remove that nasty Guava dependency.

core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -827,16 +847,29 @@ class Executor[S <: Scheduling] private[cuttle] (val platforms: Seq[ExecutionPla
...
+                    def writeln(str: CharSequence) = ExecutionStreams.writeln(nextExecutionId, str)
+                  }
+
+                  // wrap the execution context so that we can register the name of the thread of each
+                  // runnable (and thus future) that will be run by the side effect.
+                  val sideEffectExecutionContext = SideEffectExecutionContext.wrap(runnable => new Runnable {
+                      override def run(): Unit = {
+                        val tName = Thread.currentThread().getName
+                        Executor.threadNamesToStreams.put(tName, streams)
+                        runnable.run()
```
guillaumebort: Just adding a `try { ... } finally { ... }` block here is enough to always clean the map and to not have to think about potential memory leaks.

core/src/main/scala/com/criteo/cuttle/Executor.scala
BenoitHanotte: Only the thread name is exposed in the log4j LoggingEvent

core/src/main/scala/com/criteo/cuttle/Executor.scala
BenoitHanotte: Definitely more thinking is required to make sure that there would not be any leak.Because this is a quick and dirty poc to start a discussion I went the safe way with what I already knew, hence guava

core/src/main/scala/com/criteo/cuttle/Executor.scala
BenoitHanotte: I  am not sure this would be sufficient, I am mostly tthinking of InterruptedException, as the runnable handling the registration/unregistration itself is also subject to being interrupted as well as the runnable being wrapped around which we would catch the exception.
Granted this is an edge case and should likely not be an issue

BenoitHanotte:
@guillaumebort I don't think we should add the log4j part in the main cuttle module but that was required to be able to show a full example in this PR.

core/src/main/scala/com/criteo/cuttle/Executor.scala
BenoitHanotte: I updated the code

dufrannea:
Move to https://github.com/criteo/cuttle/pull/302
Thanks !

PR #291, opened 2018-07-11
Fix a bug which caused backfill states not to be persisted in the state DB

Masuzu:
The initial version of Cuttle did not cancel concurrent running executions.
If any, the scheduler would wait for them to finish before starting the backfills.

Miscellaneous:
Fixed an issue which would cause the backfills not to be registered in the state DB.

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
@@ -396,6 +396,9 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
     * Launch backfills on a set of jobs on each subperiod that can be backfilled.
     * A job has to succeed at least once on a period to be backfillable on that same period.
     * The current executions of jobs with periods overlapping the backfills are cancelled.
+    *
+    * @returns a side-effect performing the requested backfill on success, an error otherwise
+    *
     * @example Assume a backfill is requested over the period t1, t5 for the jobs = {job1, job2}
     *          For illustrative purposes, let the past execution state be as follows:
     *          <table>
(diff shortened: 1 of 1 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -434,27 +434,24 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
+      (errors.toList, validBackfills)
     }
 
-
-
     if (backfillErrors.nonEmpty)
       IO.pure(Left(backfillErrors.mkString("\n")))
-    else
+    else {
+      val executionsToCancel = listExecutionsToCancelGivenBackfills(backfillsTodo, _state(), runningExecutions.keys)
```
eryshev: Don't you think that this block should be wrapped in an atomic because it's modifying the state?

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -420,41 +420,39 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
 
-
-
     if (backfillErrors.nonEmpty)
       IO.pure(Left(backfillErrors.mkString("\n")))
-    else
+    else {
+      val (concurrentExecutionsToCancel, allExecutionCancelledPromise) = atomic { implicit txn =>
+        val executionsToCancel = listExecutionsToCancelGivenBackfills(validBackfills, currentJobStates, runningExecutions.keys)
+        (executionsToCancel, Future.sequence(executionsToCancel.map(runningExecutions.apply)))
```
eryshev: Here you try to cancel executions that you've constructed from state recuperated in a separate atomic block, it means that `executionsToCancel.map(runningExecutions.apply)` will throw if executionToCancel isn't runing anymore. And this will fail completely the Future.Sequence. Either treat this case with callbacks for each future either move this code in the same atomic as the validation. Personaly I don't think this is a good idea to separate the validation and cancellation code into two blocks

Masuzu: The 'then' following the successful (or not as you pointed out) cancellation of executions will run no matter what (as we would like it to happen: don't block a backfill because we tried to cancel a still running execution which just finished).
I get your point indeed and will refactor the code to put everything into a same synchronization block.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -420,41 +420,48 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
+    result match {
+      case Left(errors) => IO.pure(Left(errors.mkString("\n")))
+      case Right((backfills, concurrentExecutions, concurrentExecutionsPromise)) =>
+        concurrentExecutions.foreach { e =>
+          logger.info (s"Cancelling running execution ${e.id} for job ${e.job.id}")
+          e.cancel ()
+        }
+        concurrentExecutionsPromise.andThen { case _ => backfills.foreach { newBackfill =>
+          updateBackfillState (newBackfill)
+          Database.createBackfill (newBackfill).transact (xa)
```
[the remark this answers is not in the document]
Masuzu: Whoa I am not sure how this could be missed in the initial review, you are absolutely right.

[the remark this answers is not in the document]
Masuzu: Done

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -420,41 +420,48 @@ case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] wit
...
+    result match {
+      case Left(errors) => IO.pure(Left(errors.mkString("\n")))
+      case Right((backfills, concurrentExecutions, concurrentExecutionsPromise)) =>
+        concurrentExecutions.foreach { e =>
+          logger.info (s"Cancelling running execution ${e.id} for job ${e.job.id}")
+          e.cancel()
+        }
+        concurrentExecutionsPromise.andThen { case _ => backfills.foreach { newBackfill =>
+          updateBackfillState(newBackfill)
+          Database.createBackfill(newBackfill).transact(xa).unsafeRunSync
```
eryshev: Generally we don't use unsafeRunSync everywhere, try to construct an IO flow with `map/flatMap` or `IO.fromFuture` and return this constructed IO. Btw you can catch some exceptions with database interactions.

Masuzu: We can check this together the next week.
Aside from this cancelling concurrent executions when requesting a backfill should not be done imo (if there is such a need then the user should do it). I will remove the cancellation part.

PR #292, opened 2018-07-16
Initialize job success/failure metrics at 0

epsy:
Without being initialized, the metric will start at 1 success and/or failure.
If using, e.g. increase() in prometheus, it will consder 1 to be the initial value and won't have a 0 before to measure an increase from.
This means silently disappearing lone failures on monitoring boards.


Masuzu:
[For another review?] We should maybe have the `updateFinishedExecutionCounters` status parameter made into an enumeration to avoid potential issues later on when other status are added.
Like 
```
object FinishedExecutionStatus extends Enumeration {
  type FinishedExecutionStatus = Value
  val Success, Failure = Value
}

//...

FinishedExecutionStatus.values
```

epsy:
Done

epsy:
Works for me

PR #299, opened 2018-09-17
Version support

MaximeKestemont:
This adds version support in Cuttle. It allows users to know which version of Cuttle (and the jobs packaged with it) was used for each execution (and thus, the data generated by those executions). 
Changes are visible in 3 places : 
- In the UI, in the ExecutionLog (in the context object)
- In the UI, in the calendar view (through the use of a checkbox, displaying a color for each executed job + a legend mapping colors to versions)
- In the db, in the Executions (in the serialized context object)

Common use cases are :
- Know which executions needs to be backfilled in case of an bugged version
- Better understand some errors (typically, if a newly released version is trying to use data generated with an old and non backward-compatible version)
- Better trace which version of the code produced which data

On the implementation side, 
- It stores the cuttle version used by the execution in the `Context` (which is serialized in db)
- In the `Executor`class, it replaces the `projectName` by the `projectVersion`. The project name was not used in the code.
- In the UI, it adds a checkbox in the CalendarView, with the associated display logic.


timeseries/src/main/javascript/app/pages/CalendarFocus.js
```diff
@@ -219,22 +249,45 @@ const drawJobPeriods = (
     .attr("fill", "#bb65ca")
     .attr("x", MARGIN)
     .attr("y", ROW_HEIGHT - 3);
+
+    // If the version box is checked, draw a color rectangle for each done job
+    newPeriodSlot
```
Masuzu: [OPT] spacing

Masuzu:
Looking good to me, this will be convenient.

PR #306, opened 2018-10-01
Centralize and expose thread pool instantiation

Masuzu:
Others: make the scheduler checking locks on the cuttle state DB create daemon threads

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/ThreadPools.scala
+++ b/core/src/main/scala/com/criteo/cuttle/ThreadPools.scala
@@ -1,6 +1,7 @@
 package com.criteo.cuttle
 
-import java.util.concurrent.Executors
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.{ExecutorService, Executors, ScheduledExecutorService, ThreadFactory}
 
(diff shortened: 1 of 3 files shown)


Masuzu:
Test

core/src/main/scala/com/criteo/cuttle/ThreadPools.scala
```diff
@@ -13,8 +14,8 @@ object ThreadPools {
   }
 
   sealed trait Metrics {
-    protected var _threadPoolSize: Long = 0
-    def threadPoolSize(): Long = _threadPoolSize
+    protected var _threadPoolSize: AtomicInteger = new AtomicInteger(0)
```
dufrannea: no need for var here. 
Also, I think thread naming and counting is definitely an improvement, but I would like to avoid having a shared mutable thread counter. I'd rather leave it an implementaton detail of the ThreadPoolFactory

Masuzu: Oops yes indeed about the var I must have forgotten to change it back after wrapping the POD inside an Atomic object.
Sure I'll fix this it makes more sense as you said.

PR #308, opened 2018-10-05
Fix calendar view states not being updated and not refreshed

Masuzu:


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -701,24 +703,32 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject, executor: E
               acc
           }
         val upToMidnightToday = Interval(Bottom, Finite(Daily(UTC).ceil(Instant.now)))
-        (for {
+        lazy val failingExecutionIds: Set[String] = executor.allFailingExecutions.map(_.id).toSet
```
dufrannea: I do not see any reason not to leave this inlined as it was before, could you explain ?

Masuzu: Ah sure, the exists method made the call perform a linear scan on all the failing executions. Now each check is done in O(1) time instead of O(n) where n is the number of failing executions.

dufrannea: Ok makes sense, please avoid mixing perf optimization or code style improvements with bug fixing in the same PR.

PR #324, opened 2018-11-20
Fix lastruns timeseries end point

bubblesly:


vguerci:
Looks like you should rebase on master

PR #325, opened 2018-11-20
Fix the lastruns timeseries endpoint

bubblesly:
Since the version number has been added to the job states, we get
multiple intervals for each job and we can't use the fisrt we get anymore.
We merge all the intervals to retrieve their union and discard the version
in this case.

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
@@ -760,6 +760,8 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject, executor: E
           case Done(_) => true
           case _ => false
         }}
+        .foldLeft(IntervalMap.empty[Instant, Unit])((acc, elt) => acc.update(elt._1, ()))
+        .toList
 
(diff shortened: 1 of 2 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -760,6 +760,8 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject, executor: E
           case Done(_) => true
           case _ => false
         }}
+        .foldLeft(IntervalMap.empty[Instant, Unit])((acc, elt) => acc.update(elt._1, ()))
```
dufrannea: This probably works, but you can also write

```(scala)
    val successfulIntervalMaps = jobStates
        .filter(s => s._1.id == jobId)
        .values
        .flatMap(m => m.map({
          case Done(_) => true
          case _ => false
        }).toList)
```
taking advantage of the fact IntervalMap[Instant, _] is a Functor, we can map from IntervalMap[Instant, JobState] to IntervalMap[Instant, Boolean] just by providing the function 

```
jobState => jobState match {
  case Done(_) => true
  case _            => false
}
```

PR #326, opened 2018-11-26
Add tooltip for backfill priority field

Masuzu:


timeseries/src/main/javascript/app/pages/BackfillCreate.js
```diff
@@ -368,6 +379,10 @@ const styles = {
       /* Microsoft Edge */
       color: "#aaa"
     }
+  },
+  icon: {
+    fontSize: "22px",
+    marginRight: "5px"
```
eryshev: Do we really need marginRight there? Could it be done w/out or at least in relative units?

Masuzu: I'll remove the margin which is not used indeed since the icon will be placed on the right side of the component it complements already.
As for the fontSize imho that's fine since it's the size used globally in other components with icons.

eryshev: LGTM

PR #328, opened 2018-11-27
[WIP] MVP of Cron Scheduler

eryshev:
Loop for each registered job:

1 Compute next CRON event for this job (based on the CRON expression)
2 Wait for the CRON event to occur
3 Create a CronContext and run the job side-effect
4 Wait on the side-effect Future result
5 If Failure, go to 3 if we want to retry, otherwise go to 1

TODO:
- ui isn't implemented yet
- configure the number of retries per job
- persist scheduler state to DB

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,8 @@ node_modules
 dev.config.json
 .DS_Store
 
-# prices.log gererated by HelloCronScheduling
-prices.log
+# price.log gererated by HelloCronScheduling
+price.log
+
+# vi temp files
+*.un~
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,3 @@ dev.config.json
 
 # price.log gererated by HelloCronScheduling
 price.log
-
-# vi temp files
-*.un~
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ node_modules
 .idea
 dev.config.json
 .DS_Store
+
+# prices.log gererated by HelloCronScheduling
+prices.log
--- a/cron/src/main/scala/com/criteo/cuttle/cron/CronModel.scala
+++ b/cron/src/main/scala/com/criteo/cuttle/cron/CronModel.scala
@@ -1,0 +1,56 @@
+package com.criteo.cuttle.cron
+
+import java.time.temporal.ChronoUnit
+import java.time.{Duration, Instant, ZoneOffset}
+
+import com.criteo.cuttle.{Job, Scheduling, SchedulingContext, Workload}
+import cron4s.{Cron, _}
+import cron4s.lib.javatime._
+
+import scala.concurrent.duration._
+
+private[cron] case class ScheduledAt(instant: Instant, delay: FiniteDuration)
+
+/** A [[CronContext]] is passed to [[com.criteo.cuttle.Execution executions]] initiated by
+  * the [[CronScheduler]].
+  */
+private[cron] case class CronContext(instant: Instant)(retryNum: Int) extends SchedulingContext {
+  val retry: Int = retryNum
+
+  def compareTo(other: SchedulingContext): Int = other match {
+    case CronContext(otherInstant) =>
+      instant.compareTo(otherInstant)
+  }
+}
+
+/** Configure a [[com.criteo.cuttle.Job job]] as a [[CronScheduling]] job.
+  *
+  * @param cronExpression Cron expression to be parsed by https://github.com/alonsodomin/cron4s.
+  *                       See the link above for more details.
+  */
+case class CronScheduling(cronExpression: String) extends Scheduling {
+  override type Context = CronContext
+  // https://www.baeldung.com/cron-expressions
+  // https://www.freeformatter.com/cron-expression-generator-quartz.html
+  private val cronExpr = Cron.unsafeParse(cronExpression)
+
+  private def toZonedDateTime(instant: Instant) =
+    instant.atZone(ZoneOffset.UTC)
+
+  def nextEvent(): Option[ScheduledAt] = {
+    val instant = Instant.now()
+    cronExpr.next(toZonedDateTime(instant)).map { next =>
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -4,3 +4,9 @@ node_modules
 .idea
 dev.config.json
 .DS_Store
+
+# price.log gererated by HelloCronScheduling
+price.log
+
+# vi temp files
+*.un~
```
vguerci: OPT: this is clearly very specific to your env, I would suggest you to put this in your personal global git ignore `core.excludeFiles` instead

cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
```diff
@@ -0,0 +1,155 @@
...
+          logger.debug(s"Sending job ${job.underlying.id} to executor")
+          val cronContext = CronContext(scheduledAt.instant)(retryNum)
+          executor.run(job.underlying, cronContext)
+        }
+        _ <- addNextExecutionToState(job, runInfo._1)
+        _ <- logState()
+        result <- IO.fromFuture(IO(runInfo._2))
+      } yield result
+
+    def run(job: CronJob): IO[Completed] = {
```
guillaumebort: I think there is a lot of duplication between `runScheduledJob` and `run`. I would refactor your code to keep a single more linear loop here. Something like:

```suggestion
def run(job: CronJob): IO[Unit] =
  for {
    maybeScheduledAt <- IO.pure(job.nextEvent)
    _ <-
      maybeScheduledAt match {
        // No next event. Stop the loop for this job
        case None =>
          logger.info(s"Job ${job.underlying.id} has finished. We will not submit executions anymore")
          IO.unit

        case Some(scheduledAt) =>

          // Launch an execution until success or retryNum is greater than maxRetry
          def runAndRetry(retryNum: Int): IO[Completed] =
            (for {
              runInfo <- IO {
                logger.debug(s"Sending job ${job.underlying.id} to executor")
                val cronContext = CronContext(scheduledAt.instant)(retryNum)
                executor.run(job.underlying, cronContext)
              }
              _ <- addNextExecutionToState(job, runInfo._1)
              _ <- logState()
              completed <- IO.fromFuture(IO(runInfo._2))
            } yield completed).recoverWith {
              case e: Throwable =>
                if(retryNum < maxRetry) {
                  val nextRetry = retryNum + 1
                  logger.debug(s"Job ${retryJob.underlying.id} has failed it's going to be retried. Retry number: $nextRetry")
                  runAndRetry(nextRetry)
                }
                else {
                  logger.debug(s"log stop retrying...")
                  IO.raiseError(e)
                }
            }

          logger.debug(s"Run job ${newJob.underlying.id} at ${scheduledAt.instant} with a delay of ${scheduledAt.delay}")
          for {
             _ <- addNextEventToState(job, scheduledAt.instant)
             _ <- logState()
             _ <- Timeout.applyF(scheduledAt.delay)
             _ <- runAndRetry(0)
             _ <- removeJobFromState(job)
          } yield ()
      }
  } yield ()
```

I'm not sure of this code though, it's just to give you an idea.

cron/src/main/scala/com/criteo/cuttle/cron/CronModel.scala
```diff
@@ -0,0 +1,56 @@
...
+      instant.compareTo(otherInstant)
+  }
+}
+
+/** Configure a [[com.criteo.cuttle.Job job]] as a [[CronScheduling]] job.
+  *
+  * @param cronExpression Cron expression to be parsed by https://github.com/alonsodomin/cron4s.
+  *                       See the link above for more details.
+  */
+case class CronScheduling(cronExpression: String) extends Scheduling {
```
guillaumebort: Se we are missing `maxRetry` in the scheduling configuration

cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
```diff
@@ -0,0 +1,155 @@
...
+        .recover {
+          case e =>
+            val message = s"Fatal error Cron loop for job ${job.underlying.id} will be stopped"
+            logger.error(message)
+            throw new Exception(message)
+        }
+    }
+
+    val programs = workload match {
+      case CronWorkload(jobs) => jobs map New map run
```
guillaumebort: I'm sorry that's not very idiomatic regarding the rest of the code base.
```suggestion
      case CronWorkload(jobs) => jobs.map(New).map(run)
```

.gitignore
eryshev: Thanks for that, didn't know that it's considered as a bad practice.

cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
eryshev: Got it

cron/src/main/scala/com/criteo/cuttle/cron/CronModel.scala
eryshev: added it

cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
eryshev: done

cron/src/main/scala/com/criteo/cuttle/cron/CronModel.scala
```diff
@@ -0,0 +1,56 @@
...
+  override type Context = CronContext
+  // https://www.baeldung.com/cron-expressions
+  // https://www.freeformatter.com/cron-expression-generator-quartz.html
+  private val cronExpr = Cron.unsafeParse(cronExpression)
```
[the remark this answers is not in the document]
eryshev: exception is rather clear, I'm gonna add some tests for it

PR #331, opened 2018-11-29
[Bugfix] Jobs todo never launched

Masuzu:
The bug was caused by the incorrect resolution of completed parent dependencies
which spanned over contiguous time intervals, which were treated as disjoint
intervals intead of a single interval.


vguerci:
Could you please `scalafmt` code? _(Make sure you install the pre-commit hook)_

PR #343, opened 2019-01-15
Cron scheduler UI

eryshev:
MVP of an UI of Cron scheduler.
    
    It contains:
    - home page with 2 lists of active and paused jobs
    - each job can be paused or resumed
    - list of executions page for each job that shows 20 recent executions by
    default
    - for each executions we can see execution streams

Other changes:
- the system property to configure the number of threads for Cron moved to cron package


cron/index.html
```diff
@@ -0,0 +1,45 @@
+
+      
+      <h3>Active</h3>
```
guillaumebort: Is it expected to be committed?

eryshev: nope

eryshev:
> Also in term of UI I would merge both Active & Paused table (The state could be marked as "Paused" for paused jobs and the Action column could switch between Pause & Resume buttons).

Done

PR #350, opened 2019-01-25
make user's logger used everywhere

eryshev:
- we had a global implicit logger that sometimes was used instead of
user specified one, now it's deleted and user's logger persisted
everywhere

- simple implicit logger was moved to examples package

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/package.scala
+++ b/core/src/main/scala/com/criteo/cuttle/package.scala
@@ -4,7 +4,9 @@ import scala.concurrent._
 
 import cats.effect.IO
 import cats.free._
-import doobie.imports._
+import io.circe.Json
+import io.circe.parser._
+import doobie._
 
 package cuttle {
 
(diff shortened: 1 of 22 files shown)


core/src/main/scala/com/criteo/cuttle/package.scala
```diff
@@ -61,12 +63,7 @@ package object cuttle {
...
 
-  /** Default implicit logger that output everything to __stdout__ */
-  implicit val logger = new Logger {
-    def logMe(message: => String, level: String) = println(s"${java.time.Instant.now}\t${level}\t${message}")
-    override def info(message: => String): Unit = logMe(message, "INFO")
-    override def debug(message: => String): Unit = logMe(message, "DEBUG")
-    override def warning(message: => String): Unit = logMe(message, "WARNING")
-    override def error(message: => String): Unit = logMe(message, "ERROR")
-  }
+  implicit val JsonMeta: Meta[Json] = Meta[String].imap(x => parse(x).fold(e => throw e, identity _))(
```
dufrannea: either adapt name to something more self-explanatory, or add a simple comment above. (You can also do both :) )

PR #354, opened 2019-01-28
Delete timeseries_state older than 1 week.

guillaumebort:


timeseries/src/main/scala/com/criteo/cuttle/timeseries/Database.scala
```diff
@@ -40,14 +40,46 @@ private[timeseries] object Database {
...
+      WHERE date > (NOW() - INTERVAL 1 WEEK);
+
+      TRUNCATE TABLE timeseries_state;
+
+      INSERT INTO timeseries_state
+      SELECT
+        *
+      FROM timeseries_state_temp;
+
+      SELECT * from timeseries_state;
```
dufrannea: Can be removed

PR #355, opened 2019-01-30
Fixes refacto bug

dufrannea:
Number of hours and aggregationfactor were swapped.


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -150,7 +150,7 @@ object TimeSeriesCalendar {
 
 private[timeseries] object TimeSeriesCalendarView {
   def apply(calendar: TimeSeriesCalendar) = calendar match {
-    case TimeSeriesCalendar.NHourly(h)           => new NHourlyView(h, 1)
+    case TimeSeriesCalendar.NHourly(h)           => new NHourlyView(1, h)
```
eryshev: It seems strange that you're not passing timezone here.
I highly doubt that anybody is using anything different from UTC, but just in case, maybe we could add it?

PR #357, opened 2019-02-01
Configurable logs & timeseries state retention

guillaumebort:
Refactoring how the TimeSeriesScheduler is instantiated (it was
a typeclass on Scheduling before but now we don't need to retrieve
the instance implicitely).

Also cleaning the code duplication between CuttleProject.start &
CuttleProject.build

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Database.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Database.scala
@@ -2,6 +2,7 @@ package com.criteo.cuttle
 
 import java.time._
 import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.{Duration => ScalaDuration}
 
 import scala.util._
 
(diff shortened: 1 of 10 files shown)


core/src/main/scala/com/criteo/cuttle/Database.scala
```diff
@@ -353,6 +354,15 @@ private[cuttle] case class Queries(logger: Logger) {
       (${id}, ${streams})
     """.update.run
 
+  def applyLogsRetention(logsRetention: ScalaDuration): ConnectionIO[Int] =
+    sql"""
```
dufrannea: Any reason why it is not written like this ?
```sql
DELETE es
FROM executions_streams es
JOIN executions e
ON es.id = e.id
WHERE end_time < ${Instant.now.minusSeconds(logsRetention.toSeconds)}
```

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -328,7 +325,7 @@ private[timeseries] object JobState {
   * The scheduler also allow to [[Backfill]] already computed partitions. The [[Backfill]] can be recursive
   * or not and an audit log of backfills is kept.
   */
-case class TimeSeriesScheduler(logger: Logger) extends Scheduler[TimeSeries] {
+case class TimeSeriesScheduler(logger: Logger, stateRetention: Option[ScalaDuration] = None) extends Scheduler[TimeSeries] {
```
dufrannea: We should put a lower bound on this.

core/src/main/scala/com/criteo/cuttle/Database.scala
guillaumebort: I tried but it seems not supported on Mysql. I can check again.

guillaumebort: It seems to work indeed.

PR #358, opened 2019-02-04
Fix scrolling issues in Firefox and Chrome 72+

Lordshinjo:
- Chrome 72 now adheres to the css-flexbox spec and sets a non-zero
min-height for containers, which causes issues with scrolling, see
the spec at https://drafts.csswg.org/css-flexbox/#min-size-auto
- Fix executions logs not being scrollable anymore
- Fix calendar focus and job views scrolling the whole page instead of
only the job list there are too many jobs
- Fix backfill view scrolling the whole page when there are too many
backfills

### DIFF
--- a/timeseries/src/main/javascript/app/components/StreamView.js
+++ b/timeseries/src/main/javascript/app/components/StreamView.js
@@ -107,6 +107,7 @@ const styles = {
     display: "flex",
     background: "#23252f",
     position: "relative",
+    minHeight: "0",
 
     "& ul": {
       flex: "1",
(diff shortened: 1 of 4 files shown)


eryshev:
There're some glitches on low screen resolution but I suppose there were there before your change as well.

Lordshinjo:
Yeah, a low window size causes some issues but that's nothing new.
Testing with lots of log lines / jobs / backfills is more representative of the typical use case.

Alterrien:
Hi, thanks for this contrib, it'll help a lot for us Firefox users!
Small question: do you think this fix could be ported to fatcat too? There's similar UI elements where the scroll is broken too, I don't know if we could do the same fix.

Thanks!

PR #362, opened 2019-02-13
Timeseries perf

vguerci:
- Improved job selector by bumping `react-select` version
- Improved jobs list by reducing api calls and component refreshes


examples/src/main/scala/com/criteo/cuttle/examples/HelloTimeSeriesWithManyJobs.scala
```diff
@@ -0,0 +1,40 @@
+// Example: Hello cuttle timeseries with load!
```
dufrannea: This is really useful, but should not be part of "examples" IMO. Maybe move it to tests ?

vguerci: Done, had to tweak `build.sbt` for using loop on test, also tweaked the test for having long ids / hundreds of jobs tags for other tests

PR #363, opened 2019-02-15
Post event stream

vguerci:
Cuttle use EventSource, with unfortunately does not support POST / body. Because of that, projects with many jobs selected are breaking on these endpoints, reaching Blaze GET url limit.
This PR integrates a PostEventSource that mimics EventSource through POST polling. It is used to replace EventSource in all cases that can fails because of this GET limit.

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
@@ -684,11 +684,7 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject,
           case None               => BadRequest
         }
 
-    case request @ GET at url"/api/timeseries/calendar?jobs=$jobs" =>
-      val filteredJobs = Try(jobs.split(",").toSeq.filter(_.nonEmpty)).toOption
-        .filter(_.nonEmpty)
-        .getOrElse(project.jobs.all.map(_.id))
-        .toSet
+    case request @ POST at url"/api/timeseries/calendar" => {
 
       type WatchedState = ((TimeSeriesUtils.State, Set[Backfill]), Set[(Job[TimeSeries], TimeSeries#Context)])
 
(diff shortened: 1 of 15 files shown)


vguerci:
Note that this is based on #362

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -662,31 +671,38 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject, executor: E
...
-          case None               => BadRequest
+      request
+        .readAs[Json]
+        .flatMap { json =>
+          json
+            .as[CalendarFocusQuery]
+            .fold(
+              df => IO.pure(BadRequest(s"Error: Cannot parse request body: $df")),
+              query => {
+                val jobs = Try(query.jobs.filter(_.nonEmpty)).toOption
```
dufrannea: I don't get why we need a Try here

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -482,19 +496,14 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject, executor: E
           BadRequest
       }
 
-    case request @ GET at url"/api/timeseries/calendar/focus?start=$start&end=$end&jobs=$jobs" =>
+    case request @ POST at url"/api/timeseries/calendar/focus" =>
```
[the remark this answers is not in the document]
vguerci: It is used for external dependency checks, nice catch! ;)
Added a commit to support both `POST` and `GET` on this one.

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
vguerci: Me neither, just took that from previous source.
That pattern is also used in couple places of this files, removed it on my edits though.
Also there is a global `allIds` that I used instead of recomputing `project.jobs.all.map(_.id)` each time

PR #364, opened 2019-02-15
Pause/Resume Jobs

vguerci:
During the creation of the Jobs screen, the previous Paused executions Screen has been removed from UI. It included the Resume All/Filtered jobs feature.
This PR does:
- Remove Paused executions dead code
- Moves Pause/Resume All/Filtered jobs action to the Jobs Screen


vguerci:
Note that this is based on #362 and #363

vguerci:
Added a commit to change popup menu with single pause/resume action to a simple play/pause button (one click saved!)

PR #370, opened 2019-02-22
Fix fragmented backfill creation

guillaumebort:
This was caused because we create one backfill for each interval. Now that
we track a different interval for each version it means that asking for a
backfill we often create several backfill for the same jobs for contiguous
intervals.

This fix merge them together when possible.


guillaumebort:
please review based on #375

PR #379, opened 2019-02-27
Fix posteventsource

guillaumebort:


timeseries/src/main/javascript/Utils.js
```diff
@@ -39,7 +42,7 @@ export class PostEventSource {
         .then(data => data.json())
         .then(
           data => {
-            this._observers.forEach(o => o({ data }));
+            this._polling && this._observers.forEach(o => o({ data }));
             this._poller = setTimeout(() => poll(), 5000);
```
dufrannea: Shouldn't we apply this log to the setTimeout as well ?

guillaumebort: Fixed, and also in err case

PR #381, opened 2019-03-06
Fix force success

bubblesly:
We only cancel executions that correspond to the interval parameter of the force success command

### DIFF
--- a/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
+++ b/timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
@@ -963,10 +963,12 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject,
         job <- Try(project.jobs.all.find(_.id == jobId).getOrElse(throw new Exception(s"Unknow job $jobId")))
       } yield {
         val requestedInterval = Interval(startDate, endDate)
-        //self.forceSuccess(job, requestedInterval, executor.cuttleProject.version)
         scheduler.forceSuccess(job, requestedInterval, executor.projectVersion)
-        val runningExecutions = executor.runningExecutions.filter(e => e._1.job.id == jobId).map(_._1)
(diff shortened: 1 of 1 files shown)


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -963,10 +963,12 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject,
         job <- Try(project.jobs.all.find(_.id == jobId).getOrElse(throw new Exception(s"Unknow job $jobId")))
       } yield {
         val requestedInterval = Interval(startDate, endDate)
-        //self.forceSuccess(job, requestedInterval, executor.cuttleProject.version)
         scheduler.forceSuccess(job, requestedInterval, executor.projectVersion)
-        val runningExecutions = executor.runningExecutions.filter(e => e._1.job.id == jobId).map(_._1)
-        val failingExecutions = executor.allFailingExecutions.filter(_.job.id == jobId)
+        def filterOp(e: Execution[TimeSeries]): Boolean = {
+          e.job.id == jobId && e.context.start.compareTo(startDate) > 0 && e.context.end.compareTo(endDate) < 0
```
eryshev: Why don't we use timeseries.interval.Interval.intersect for this task?

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesApp.scala
```diff
@@ -963,10 +963,12 @@ private[timeseries] case class TimeSeriesApp(project: CuttleProject,
...
+        }
+        val runningExecutions = executor.runningExecutions.map(_._1).filter(filterOp)
```
eryshev: OPT: map.filter is replaced by collect usually

eryshev:
Could you also add a more explicit commit message?

PR #426, opened 2019-07-25
introduce a time window batching

eryshev:
- user can configure a time window batching per job
- an example is added to HelloTimeSeries


timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -960,12 +988,46 @@ case class TimeSeriesScheduler(logger: Logger,
...
+      if (job2Contexts.isEmpty) {
+        logger.debug(s"nothing to run for ${job.name}")
+        atomic { implicit txn =>
+          debouncedJobs() = debouncedJobs() - job
+        }
+        List.empty
+      } else {
+        if (job.scheduling.batching.size > 1) {
+          val now = Instant.now()
+          debouncedJobs.single().contains(job) match {
```
dufrannea: - There is no concurrent access to _debouncedJobs_ (scheduling loops are serialized), so technically we do not need STM
- No side effect in _jobsToRun_, please take as an argument & return the _debouncedJobs_ from _jobsToRun_. You can side effect them in the calling method where all the rest is side-effected

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -962,10 +1009,28 @@ case class TimeSeriesScheduler(logger: Logger,
...
+
+    job2Contexts.partition {
+      case (job, _) =>
+        if (job.scheduling.batching.size == 1) {
+          true
+        } else {
+          debouncedJobs.contains(job) match {
+            case true if now.isAfter(debouncedJobs(job)) =>
+              true
+            case true =>
```
dufrannea: simplify case

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
```diff
@@ -723,16 +749,37 @@ case class TimeSeriesScheduler(logger: Logger,
...
+      case (job, _) =>
+        if (job.scheduling.batching.size > 1) {
+          logger.debug(s"scheduling batches for ${job.name}")
+          debouncedJobs -= job
+        }
+    }
+
+    debounced.groupBy(_._1).foreach {
+      case (job, _) =>
+        debouncedJobs.contains(job) match {
```
dufrannea: OPT: debouncedJobs.get() match { Some(x) => ... } to avoid the debouncedJobs(job) afterwards

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
eryshev: done

timeseries/src/main/scala/com/criteo/cuttle/timeseries/TimeSeriesScheduler.scala
eryshev: done

PR #450, opened 2019-08-09
Update lolhttp to 0.13.0

vguerci:
Also update:
- transitive dependencies: cats-core, circe, doobie
- scala 2.12.8


vguerci:
Requires lolhttp 0.13.0 to be published with the related dependencies change: criteo/lolhttp#92

PR #486, opened 2019-11-11
Cron dag

szehonCriteo:


szehonCriteo:
Adding a screenshot, in this example there are several runs of the cron-job, each with execution of each part.  In this case, the cron job is a dag with with several parts (part1 , 2, 3) 

Screenshot 2019-11-10 at 23 53 29

szehonCriteo:
Also, this does not do any checkpointing in case cuttle cron goes down.  This can be added later.

guillaumebort:
On the API design:

I think it would be better to keep cuttle `Job[CronScheduling]` instead of introducing this `CronJob`. So a cuttle job would be what you call `CronJobPart` now. Several jobs would be grouped into a `CronPipeline` which is a dag of jobs + a cron expression. The final workload would be a Set of these pipeline things.

On the UI:

It would be useful to group job executions for the same pipeline. It is probably easy to by using the CronEvent as a key.

cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
```diff
@@ -125,35 +124,54 @@ case class CronScheduler(logger: Logger) extends Scheduler[CronScheduling] {
...
+            for {
+               _ <- runAndRetry(job,
+                         node,
+                         scheduledAt,
+                         node.maxRetry,
+                         runNowUser)
+               _ <- IO(state.cronJobPartFinished(job, node))
+               result <- runNextPart(job, scheduledAt, runNowUser)
+           } yield result
+        }.toList.sequence_.map(_ => Completed)
```
guillaumebort: `sequence` here will run the IO sequentially. You can use `parSequence` instead since these steps can be parallelised: https://typelevel.org/cats-effect/datatypes/io.html#parsequence

guillaumebort:
The API would be something like this:

```scala
val doXstep1 = Job(id = "x1", scheduling = CronScheduling(maxRetry = 1),) { ... }
val doXstep2 = Job(id = "x2", scheduling = CronScheduling(),) { ... }
val doY = Job(id = "y", scheduling = CronScheduling(),) { ... }

val cronDag1 = (doXstep2 dependsOn doXstep1).every("0-59/10 * * ? * *")
val cronDag2 = doY.every("0-59/10 * * ? * *")

val workload = Set(cronDag1, cronDag2)
```

szehonCriteo:
Hi @guillaumebort  thanks for the review, addressed the suggested changes as best I could to refactor cronJob -> cronDag, cronJobPart -> cronJob.

One open question is about the suggested API .every(), because CronDag needs an id for the UI and state so I did toCronDag instead.  Open to suggestions about this.

Another open question is about the pauseJobs, right now they actually correspond to pauseDags because the UI allows user to pause the Dag and not individual jobs.  We should implement pauseJob UI endpoint, but then do we store both pausedDags and pausedJobs in the database then?  It can be in the subsequent pull request, right now it works for pauseDag.

And I also improved the UI from the previous screenshot, mostly the same but just unifying the run element.

Screenshot 2019-12-09 at 14 12 50

PR #491, opened 2019-12-03
Add default values to clean state DB

juangentile:
Default values are:

30 days for executions
1 day for state
10 max versions


juangentile:
https://github.com/criteo/cuttle/commit/5c42267d23c37bd6b28ec90fc97f171883f2d43f

PR #492, opened 2019-12-04
Add a relaunch method to relaunch specific executions

bubblesly:
Code has been factorized with relaunch which is specific to jobs

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Executor.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Executor.scala
@@ -635,10 +635,10 @@ class Executor[S <: Scheduling](val platforms: Seq[ExecutionPlatform],
   def openStreams(executionId: String): fs2.Stream[IO, Byte] =
     ExecutionStreams.getStreams(executionId, queries, xa)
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/criteo/cuttle/Executor.scala
```diff
@@ -635,10 +635,10 @@ class Executor[S <: Scheduling](val platforms: Seq[ExecutionPlatform],
   def openStreams(executionId: String): fs2.Stream[IO, Byte] =
     ExecutionStreams.getStreams(executionId, queries, xa)
 
-  def relaunch(jobs: Set[String])(implicit user: User): Unit = {
+  private def relaunchExecs(idList: Set[String], cond: (Set[String], Execution[S]) => Boolean)(implicit user: User): Unit = {
```
vguerci: I suggest keeping jobs as first argument and using predicate (or filter) as second

vguerci: or better, write a generic `private def relaunchExecutions(predicate: Execution[S] => Boolean)`

bubblesly: done thx

PR #510, opened 2019-12-23
Bump version to 0.11.0

szehonCriteo:


szehonCriteo:
Dag is incompatible API, so bumping the minor version

PR #514, opened 2020-01-27
Cron fixes

Lordshinjo:
2 main fixes:
* Any job failure would break the whole execution loop for all jobs
* Jobs would ignore configured retries by starting at their max attempt

Additionally, the UI was improved a bit and the code reformatted properly

### DIFF
--- a/cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
+++ b/cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
@@ -92,8 +92,8 @@ case class CronScheduler(logger: Logger) extends Scheduler[CronScheduling] {
     }
   }
 
-  private[cron] def resumeDags(dags: Set[CronDag],
-                               executor: Executor[CronScheduling])(implicit transactor: XA, user: Auth.User): Unit = {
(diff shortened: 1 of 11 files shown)


cron/src/main/scala/com/criteo/cuttle/cron/CronScheduler.scala
```diff
@@ -147,30 +147,33 @@ case class CronScheduler(logger: Logger) extends Scheduler[CronScheduling] {
...
+      runIO.flatMap {
+        case Left(e) =>
           if (retryNum < job.scheduling.maxRetry) {
             val nextRetry = retryNum + 1
             logger.debug(s"Job ${job.id} has failed it's going to be retried. Retry number: $nextRetry")
-            runAndRetry(dag, job, scheduledAt, nextRetry, runNowUser)
+            runAndRetry(dag, job, scheduledAt, runNowUser, retryNum = nextRetry)
           } else {
             logger.debug(s"Job ${job.id} has reached the maximum number of retries")
-            IO.raiseError(e)
```
szehonCriteo: Wow stupid mistake, thanks for fixing it!

PR #533, opened 2020-04-14
Expose Execution.cancel() more properly

Lordshinjo:
- The current `cancel` function requiring a `User` is made internal to cuttle and renamed `userCancel`
- Add a new `cancel` function that takes no parameter, that does the same but with no logging


dufrannea:
Hi, thanks for the contribution. To be honest I am a bit reluctant to continue exposing this cancel mechanism to user, when they can just fail the execution. Here is what I suggest:

- put the current cancel() as private as you did and rename it "internalCancel" or something of the sort
- add a cancel() method visible to users that throws an exception.

What do you think ?

Lordshinjo:
Thinking again about it, any cancel mechanisms exposed to the user in the middle of an execution fit better in the `ExecutionPlatform` methods.

Putting `cancel()` as private should be enough, I'll probably do it in another PR.

PR #558, opened 2020-08-05
Clean Cron UI

Lordshinjo:
Add UI for Cron that looks like the TimeSeries UI.
This UI is still missing some things that were already missing from the old UI, such as the visualization of job dependencies inside a CronDag, but it is already more usable than the old UI.

This also fixes some things in the Cron Scheduler, such as the CronContext which was unique per job retry, making retries ignore the RetryStrategy and retry immediately every time.

To avoid duplicating too many thing, the UI for both TimeSeries and Cron has been moved in a new folder containing both UIs.


Lordshinjo:
Splitting in 2 PRs

PR #604, opened 2021-03-18
[data-catalog] create cuttle "readOnlyConnect" method

manudizandrei:
In the read only connection, the db lock is bypassed and the transactor is
returned as is.

JIRA: DPT-2819

### DIFF
--- a/core/src/main/scala/com/criteo/cuttle/Database.scala
+++ b/core/src/main/scala/com/criteo/cuttle/Database.scala
@@ -223,6 +223,14 @@ object Database {
       }
     )
   }
+
+  /* To be used only with a read-only db user as the mandatory lock check / creation is bypassed. */
+  def readOnlyConnect(dbConfig: DatabaseConfig)(implicit logger: Logger): XA = {
+    // FIXME we shouldn't use allocated as it's unsafe instead we have to flatMap on the Resource[HikariTransactor]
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/criteo/cuttle/Database.scala
```diff
@@ -223,6 +223,14 @@ object Database {
       }
     )
   }
+
+  /* To be used only with a read-only db user as the mandatory lock check / creation is bypassed. */
+  def readOnlyConnect(dbConfig: DatabaseConfig)(implicit logger: Logger): XA = {
+    // FIXME we shouldn't use allocated as it's unsafe instead we have to flatMap on the Resource[HikariTransactor]
+    val (transactor, _) = newHikariTransactor(dbConfig).allocated.unsafeRunSync
+    logger.debug("Allocated new read-only Hikari transactor")
```
antonlin1, line 231: could we change this to `logger.info` instead?

antonlin1 (changes requested):
just minor on log-level

core/src/main/scala/com/criteo/cuttle/Database.scala
manudizandrei: sure!

manudizandrei: updated it

antonlin1 (approved):

